PR #39, opened 2015-02-11
Package config command change for Laravel 5

RMFogarty:


tymondesigns:
Thanks, but I intend to finish up the wiki, and scrap the current readme as it is for version `0.3.*`

PR #43, opened 2015-02-12
Update to vendor:publish Laravel 5

angelmartz:
Update to vendor:publish instead config:publish


tymondesigns:
Thanks, but I intend to finish up the wiki, and scrap the current readme as it is for version 0.3.*

PR #38, opened 2015-02-09
Add multidimensional custom claim support

JDqv:
Had a need for multidimensional custom claim support, so I tweaked jwt-atuh a bit.

With this payload accepts claims containing arrays.

PR #45, opened 2015-02-17
Add multidimensional custom claim support

JDqv:
Rebased #38


tymondesigns:
Thanks for this, I have implemented it and fixed tests since it was a quick change

PR #78, opened 2015-04-06
Bug Fix for Refresh Token middleware raising 'Token is Blacklisted' erro...

manadan999:
...r.

It was creating a new token and invalidating the current request token before processing the request. So by the time 'after middleware' kicks in, its current token is already invalid. This fix is all it needs.

### DIFF
--- a/src/Middleware/RefreshToken.php
+++ b/src/Middleware/RefreshToken.php
@@ -16,15 +16,16 @@ class RefreshToken extends BaseMiddleware
      */
     public function handle($request, \Closure $next)
     {
-        try {
+
+        $response = $next($request);
+        
+         try {
             $newToken = $this->auth->setRequest($request)->parseToken()->refresh();
         } catch (TokenExpiredException $e) {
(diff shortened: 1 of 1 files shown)


tymondesigns:
Could you resolve the formatting as per:
https://styleci.io/commits/05af04bf703664c82953d15050a9eb66db07e632
.. and I will merge

Thanks

tymondesigns:
added 9dfebcb36ebf3268794b7e461549e512f16dd222

Maykonn:
Hi!

I update the RefreshToken middleware with propose fix(composer update not update jwt-auth 0.5.*), and I'm getting the following error on next request with a refreshed token:
```
{
    "error": "token_invalid"
}
```

I ask for help on Stackoverflow, just to share more information about this awesome lib:
http://stackoverflow.com/questions/29703260/laravel-jwt-tokens-are-invalid-after-refresh-them-in-a-authentication-jwt-approa

PR #99, opened 2015-04-21
clone new jwt provider without auth for lumen

devmark:
app.php register:
~~~~
$app->register('Tymon\JWTAuth\Providers\JWTServiceProvider');
~~~~
use:
~~~~
use Tymon\JWTAuth\Facades\JWT;
$payload = app('tymon.jwt.payload.factory')->sub('id')->aud('foo')->make();
$token = JWT::encode($payload);
~~~~

middleware
~~~~
...
    public function __construct(ResponseFactory $response, Dispatcher $events, JWT $jwt)
    {
        $this->jwt = $jwt;
        $this->response = $response;
        $this->events = $events;
    }
...
    public function handle($request, \Closure $next)
    {
        if (!$token = $this->jwt->setRequest($request)->getToken()) {
            return $this->respond('tymon.jwt.absent', 'token_not_provided', 400);
        }

        try {
            $payload = $this->jwt->decode($token);
             //your auth
        } catch (TokenExpiredException $e) {
            return $this->respond('tymon.jwt.expired', 'token_expired', $e->getStatusCode(), [$e]);
        } catch (JWTException $e) {
            return $this->respond('tymon.jwt.invalid', 'token_invalid', $e->getStatusCode(), [$e]);
        }
...
~~~~

### DIFF
--- a/src/Facades/JWT.php
+++ b/src/Facades/JWT.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Tymon\JWTAuth\Facades;
+
+use Illuminate\Support\Facades\Facade;
+
+class JWT extends Facade
+{
+  /**
+   * Get the registered name of the component.
+   *
+   * @return string
+   */
+  protected static function getFacadeAccessor()
+  {
+    return 'tymon.jwt';
+  }
+}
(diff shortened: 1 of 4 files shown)


tymondesigns:
Hi Mark,

I don't think I'm keen on having this much duplication in the package. Although I do see exactly what you are achieving by doing so.

I am currently considering splitting out the core from the package at some point. So then it would be easier and simpler to have say, `jwt-auth`, `jwt-auth-laravel` and `jwt-auth-lumen` etc

Issue #101, opened 2015-04-21
how to use Token refresh in L5


ykenny:
when i got tips it says "token expire" , how can i refresh and got new token?
 i have been use JWTAuth::refresh($oldToken) but it doesn't work .

jfadich:
This is an issue with the package. In the refresh method when decode is called the token is validated automatically, but the refreshFlow defaults to false so a token expiration exception is thrown. 

I submitted a pull request that fixes the issue. You can look at it and use the changes, or wait to see if the pull request is accepted.

tymondesigns:
should be fixed as of `0.5.3`

PR #102, opened 2015-04-22
Fix issue causing auth refresh to fail

jfadich:
I had an issue where I would always get a expired token error when trying to refresh a token that is expired, but still within the refresh_ttl.

I found that the token was being validated before the refreshFlow was enabled. This pull request adds the ability to start the refresh flow before validating.

Another user encountered the same problem and created issue #101

### DIFF
--- a/src/JWTManager.php
+++ b/src/JWTManager.php
@@ -29,6 +29,11 @@ class JWTManager
     protected $blacklistEnabled = true;
 
     /**
+     * @var boolean
+     */
+    protected $refreshFlow = false;
+
+    /**
      *  @param \Tymon\JWTAuth\Providers\JWT\JWTInterface  $jwt
      *  @param \Tymon\JWTAuth\Blacklist  $blacklist
      *  @param \Tymon\JWTAuth\PayloadFactory  $payloadFactory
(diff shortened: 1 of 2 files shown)


src/JWTManager.php
```diff
@@ -81,7 +86,7 @@ public function decode(Token $token)
      */
     public function refresh(Token $token)
     {
-        $payload = $this->decode($token);
+        $payload = $this->setRefreshFlow()->decode($token, true);
```
tymondesigns: second param isn't needed

tymondesigns:
Thanks for this.. please see my comment on commit

tymondesigns:
Also it would be great if you could squash your commits :+1:

jfadich:
Whoops, good catch. That was a remnant from before my refactor. I'll fix that and squash my commits tonight when I get home from work.

jfadich:
I've updated this pull request

jfadich:
not sure why that test failed. It passed on my end with php 5.5.12

tymondesigns:
It sometimes happens, weirdly.. I think it may have something to do with using the actual time() function in the tests, and Travis running a little slower, so the mocked values don't always match up.

I think I will remove those refs at some point, to see if it fixes it.

Thanks

PR #108, opened 2015-04-30
[WIP] Added support for closures as custom claims

### LINKED ISSUE #89: Adding multiple columns as custom claims
Is it possible to add multiple model columns as claims ?
I would like to add a second column from the model to the token payload (for instance: "role")
I found a post here: http://stackoverflow.com/questions/29014681/add-the-user-role-to-the-jwt-laravel-5-jwt-auth but I don't like the idea of hitting the database twice, is there another way of achieving this ?

rscafi:
As [my suggestion](https://github.com/tymondesigns/jwt-auth/issues/89#issuecomment-97771027) to fix #89, it makes possible to pass closures as Custom Claims, with the User as the first argument:

```php
$customClaims = [
    'regular' => 'A regular claim',
    'role' => function($user) {
        return $user->role;
    },
];

$token = JWTAuth::attempt($credentials, $customClaims);
```

## Work in progress:
- [X] Basic support
- [X] Support for callables
- [ ] Write some tests

### DIFF
--- a/src/JWTAuth.php
+++ b/src/JWTAuth.php
@@ -81,7 +81,7 @@ class JWTAuth
      */
     public function fromUser($user, array $customClaims = [])
     {
-        $payload = $this->makePayload($user->{$this->identifier}, $customClaims);
+        $payload = $this->makePayload($user, $customClaims);
 
         return $this->manager->encode($payload)->get();
     }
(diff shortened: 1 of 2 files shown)


tymondesigns:
thank you for starting this, however I have gone a different direction with it... as you can see [here](https://github.com/tymondesigns/jwt-auth/tree/develop)

santyanna:
Is it possible to elaborate on the solution to this @tymondesigns? Unfortunately your link just goes to the develop branch and not specifically how to solve this problem.

PR #121, opened 2015-05-22
Permit to pass custom claims to the refresh method

boudydegeer:
I need to change the is claim also on the refresh method and it was to complicated, by passing the custom claims directly to the method as second parameter its possible.


tymondesigns:
Thanks for highlighting this... but it will come with the next release

Issue #117, opened 2015-05-18
jwt.auth middleware - returning status code 400 on TokenExpiredException


Neoglyph:
Hey everyone, 

I have stumbled across a problem, where the jwt-auth middleware (the ```Tymon\JWTAuth\Middleware\GetUserFromToken``` class) on an ```ExpiredTokenException``` would keep returning me a status code of 400, instead of the correct 401, all while returning the correct response - ```["token_expired"]```.

I have got it working, by adding namespaces before the Exceptions, which seemed to return the correct $e->getStatusCode() and fixed the issue.

I just wanted to post this here, since it will maybe be of use to someone, who might stumble on the same problem, that I have.

``` php
namespace Tymon\JWTAuth\Middleware;

class GetUserFromToken extends BaseMiddleware
{
    ...
    public function handle($request, \Closure $next)
    {
        ...
        /* before - returns incorrect $e->getStatusCode() */
        try {
            $user = $this->auth->authenticate($token);
        } catch (TokenExpiredException $e) {
            return $this->respond('tymon.jwt.expired', 'token_expired', $e->getStatusCode(), [$e]);
        } catch (JWTException $e) {
            return $this->respond('tymon.jwt.invalid', 'token_invalid', $e->getStatusCode(), [$e]);
        }
        /* after - returns correct $e->getStatusCode() */
        try {
            $user = $this->auth->authenticate($token);
        } catch (Tymon\JWTAuth\Exceptions\TokenExpiredException $e) {
            return $this->respond('tymon.jwt.expired', 'token_expired', $e->getStatusCode(), [$e]);
        } catch (Tymon\JWTAuth\Exceptions\JWTException $e) {
            return $this->respond('tymon.jwt.invalid', 'token_invalid', $e->getStatusCode(), [$e]);
        }
       ...
    }
}
```

rathesDot:
Had the same problem and fixed it the same way as @Neoglyph described. Works well and should be updated.

jdsingh:
@Neoglyph Had the same problem. Status code should be 401 for this package.

It returns 404 for user_not_found where it should return 401 when using `GetUserFromToken` middleware.

jasanmiguel:
I have the same problem. Status should be 401

npx:
Can we keep this open as this is not fixed in 0.5.3?

Mr-Anonymous:
[...]

PR #135, opened 2015-06-03
status code on error should be 401

jdsingh:
fixes issue #117

### DIFF
--- a/src/Exceptions/InvalidClaimException.php
+++ b/src/Exceptions/InvalidClaimException.php
@@ -7,5 +7,5 @@ class InvalidClaimException extends JWTException
     /**
      * @var integer
      */
-    protected $statusCode = 400;
+    protected $statusCode = 401;
 }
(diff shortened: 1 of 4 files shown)


codeorganic:
User Not Found should be 404 @jdsingh 

This pull request should be rejected

jdsingh:
@codeorganic Purpose of this middleware is to find the user by auth token and log user into the app. 

If the user account does not exists that makes the Request invalid and code 404 should be user where the request is valid but the resource requested does not exist.

Check definition of [404](http://httpstatus.es/404)  it says "Subsequent requests by the client are permissible." but in this case they are not.

tymondesigns:
I have now deleted the `laravel-5` branch, I will add this in myself, unless you want to recreate it targetting `develop`

Mr-Anonymous:
Is this fixed? I installed it on Laravel 5.1 and I noticed the same. For token expired errors, the auth middleware actually returns `{"error":"token_not_provided"}` with a 400 status. Due to which the http interceptor are not triggered.

PR #116, opened 2015-05-17
Upgrade to support Doctrine entities

oscaragcp:
Doctrine entitiy attributes are not usually public, so it is necessary to access them by a method instead of directly. I have made a small fix to support this without having to change anything else.

PR #145, opened 2015-06-10
Upgrade to support Doctrine entities

oscaragcp:
Doctrine entitiy attributes are not usually public, so it is necessary to access them by a method instead of directly. I have made a small fix to support this without having to change anything else.


tymondesigns:
`0.6` will come soon #116

PR #166, opened 2015-07-03
Update PayloadFactory

### LINKED ISSUE #165: exp, nbf and iat change value from string to int
Hi,
I have a problem with the given claims: exp,nbf,iat. All these values are returned as a string.
Is there a way to change the values from a string to an integer value??

Update: read the specs here https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.4

and all these claims should be an integer!!

Thanks

lkoenigsberger:
Updated return values to integer for iat,exp,nbf claims
Closes #165

### DIFF
--- a/src/PayloadFactory.php
+++ b/src/PayloadFactory.php
@@ -155,7 +155,7 @@ class PayloadFactory
      */
     public function iat()
     {
-        return Utils::now()->format('U');
+        return (int) Utils::now()->format('U');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tymondesigns:
Thanks

stevenmyhre:
Any chance you'll have a stable/tagged release soon that will include this? It's blocking some interoperability with some of my node server libraries that require these claims be integers.

tymondesigns:
@stevenmyhre not committing to a date but hopefully soon, Docs writeup underway...

Although passing `JSON_NUMERIC_CHECK` to your json response should fix this for you for now. e.g.

```php
return response()->json($data, 200, [], JSON_NUMERIC_CHECK);
```

PR #177, opened 2015-07-10
Fix #176

### LINKED ISSUE #176: JWT payload does not contain the required claims
Hey @tymondesigns, just updated and got 
> TokenInvalidException in PayloadValidator.php line 57:
> JWT payload does not contain the required claims

murwa:
I'm not sure that's the correct solution but I needed it to work

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor
 composer.lock
 build
+.idea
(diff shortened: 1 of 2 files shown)


tymondesigns:
Ahh right I understand your issue now. This was removed recently, now the correct way to do is:

```php
$payloadFactory->customClaims(['foo' => 'bar'])->make();

Murwa:
ok

kamil-kielczewski:
In file `config/jwt.php` from key

`'required_claims' => ['iss', 'iat', 'exp', 'nbf', 'jti', 'sub']`

I remove unused claims (in my case `'sub'`)

PR #204, opened 2015-08-03
Bump namshi/jose to 6.0.x

xiphiaz:
This major update of namshi/jose includes fixes to json encoding of slashes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": ">=5.4.0",
         "illuminate/support": "~5.0",
         "illuminate/http": "~5.0",
-        "namshi/jose": "5.0.*",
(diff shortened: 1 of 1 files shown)


tymondesigns:
Thankyou for highlighting this, but development is only being made on the develop branch atm

added here - 525f3293dc4eacebc6347986d940cee3460691c6

mfn:
Please that this is a possible incompatible change, which means tokens generated with JOSE 5.0 *may* not compatible with JOSE 6.0 .

If the payload contains a `/`, the verification will fail. I learned this the hard way because of using `tymon/jwt-auth` (great package, btw!) with JOSE 5  with another service end which uses raw JOSE, but version 6.0 .

May be worth documenting somewhere ...

Issue #127, opened 2015-05-27
Using refresh middleware, token is invalid


mattmcdonald-uk:
On my route I'm using the jwt.refresh middleware.

I set the ttl to 1 to make sure refresh was working correctly, and encountered an issue.

When I try to get the current user (JWTAuth::parseToken() -> toUser()) in my controller, because the token hasn't yet been refreshed the token is invalid and it throws an exception.

Another issue I encountered here was that instead of it throwing an exception that I could catch, it seems to throw an exception I can't catch - not sure why. But my error log shows: local.ERROR: exception 'Tymon\JWTAuth\Exceptions\TokenExpiredException' with message 'Token has expired

Attempts to catch TokenExpiredException don't work.

This issue occurs in 0.5.3 and in the latest dev version.

My workaround has been to modify RefreshToken so that the token is refresh before calling $next. Not sure whether this is a correct fix, or whether this might have an unwanted impact elsewhere.

codeorganic:
Yes I believe you are saying:
"->refresh()" should not just blindly invalidate a token and it should check to make sure that it needs a refresh before"  

This would fix the main issue with using ->refresh() in the current supplied middleware.

mattmcdonald-uk:
@codeorganic The issue here is that the token is already invalid, but can be refreshed. But because of the order the middleware is run in, the expired token is used to get the user before it's refreshed.

As it's an expired token, getting the user isn't possible.

osteel:
As for the non caught exception, make sure you're trying to catch `\Tymon\JWTAuth\Exceptions\TokenExpiredException` and not `Tymon\JWTAuth\Exceptions\TokenExpiredException` (mind the backslash at the beginning)

tymondesigns:
@mattmcdonald-uk just to clarify, are you applying both the middlewares to the same route, or just `RefreshToken` ?

mattmcdonald-uk:
@tymondesigns I'm just applying the refresh middleware. I originally applied both, but that didn't work as the first middleware was failing because the token was invalid (pre-refresh). Same problem the parseToken method is having.

@osteel Yes - spot on with the missing backslash (docs here also missing it: https://github.com/tymondesigns/jwt-auth/wiki/Authentication but perhaps the presume you've declared Tymon above?)

osteel:
[...]

PR #215, opened 2015-08-14
Fixes security related bug in RefreshToken middleware

schuppo:
Fixes security related bug

I was about to use the middleware as authorization middleware when I realized that it inside it is implemented as an after middleware. If I am not totally wrong that means that any request to a related route also those without authorization header will call the related controller action (and for example update or delete a resource) before the requester is told not being allowed to do so.

I didn't write any functional tests (because I didn't find any which test if unauthorized users are blocked properly from doing forbidden things) but you get the idea.

If you don't think that this issue is a problem and the RefreshToken should be kept as is, it is fine for me. But I'd suggest then to update the wiki to let users know about the risk when using the RefreshToken middleware without the GetUserFromToken middleware.

Btw. jwt-auth is a great Laravel extension

### DIFF
--- a/src/Middleware/RefreshToken.php
+++ b/src/Middleware/RefreshToken.php
@@ -1,14 +1,5 @@
 <?php
 
-/*
- * This file is part of jwt-auth
- *
- * (c) Sean Tymon <tymon148@gmail.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 namespace Tymon\JWTAuth\Middleware;
 
 use Tymon\JWTAuth\Exceptions\JWTException;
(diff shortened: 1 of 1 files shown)


tdhsmith:
I thought this too, but based on the discussion in #127, it apparently isn't the current intended use of the refresh middleware:

> I would advocate a separate endpoint for each purpose. i.e. One for retreiving the authenticated user and one for simply refreshing tokens. 

> [...]

> When a 401 response is returned (due to token expiration) then you would make a subsequent request to a refresh endpoint, to replace the expired token.

That is, it isn't currently 'supposed to' be used on routes that do anything else; it belongs on an endpoint that only does token refreshes.

Of course this means it can't really be used for a 'single-use token' set up, but it sounds like @tymondesigns is evaluating how to combine all these different approaches...

schuppo:
Makes totally sense to me. So probably it is best not to merge the current development branch with my pull request. 

But I'd still suggest to say something about the purpose of the RefreshToken middleware in the wiki, because it is not clear on first sight.

tdhsmith:
Yeah I think it might be helpful to have a little diagram of the token timelines and how the middleware behaves at each point. Might investigate that tonight...

tymondesigns:
refactoring the middleware over the next couple of days

Thanks for raising this though

mcblum:
I'm a little late to this but it'd be awesome if there was a way to easily incorporate token refreshes into middleware. Sounds like that's the direction it's going. Thanks for a great Laravel package.

PR #218, opened 2015-08-18
Fix issues with blacklisting expired tokens and refreshing blacklisted tokens

### LINKED ISSUE #143: Token Refresh / Expiry / Invalidation
Found and strange behavior!

I refresh a token with blacklist enabled witch is expired!

$token = JWTAuth::parseToken()->refresh();

Now this token should be invalidated for any further requests but since it's expired it's not added to the invalidation list, so i could use it to get a new token witch is i think a little odd

tdhsmith:
Fixes #143; blacklisting now **immediately** invalidates a token, regardless of expiry status, addressing a number of small issues:

* Expired tokens were just being ignored, preventing refreshable tokens from being blacklisted.
* Grace period was starting at a token's expiry date, so `blacklist->has()` would often return false for blacklisted tokens that hadn't reached their `exp` claim. Grace period now starts at time of blacklisting.
* Storage time had only extended until a token's expiration date, so depending on storage provider, entries may be cleared, allowing post-expiry tokens to sometimes become refreshable again. Entry storage time now extends to the expiry date or `iat + refesh_ttl`, whichever is later.
* Some fixes for expected time behavior and enforcing number of method calls to strengthen Blacklist unit tests.

This definitely needs some eyes & outside testing, but I wanted to put it up for consideration before heading home for the day. ;)  The new cases might also merit some extra unit tests.

There's probably also some important discussion to be had about whether anyone was relying on the old behavior (the grace period setting in particular?).

### DIFF
--- a/src/Blacklist.php
+++ b/src/Blacklist.php
@@ -30,6 +30,13 @@ class Blacklist
     protected $gracePeriod = 0;
 
     /**
+     * Number of minutes from issue date in which a JWT can be refreshed.
+     *
+     * @var integer
+     */
+    protected $refreshTTL = 20160;
+
+    /**
      * @param \Tymon\JWTAuth\Contracts\Providers\Storage  $storage
      */
     public function __construct(Storage $storage)
@@ -122,6 +129,20 @@ class Blacklist
     public function setGracePeriod($gracePeriod)
     {
         $this->gracePeriod = (int) $gracePeriod;
+
+        return $this;
+    }
+
+    /**
+     * Set the refresh time limit
+     *
+     * @param  integer
+     *
+     * @return Blacklist
+     */
+    public function setRefreshTTL($gracePeriod)
+    {
+        $this->refreshTTL = $ttl;
 
         return $this;
     }
--- a/src/Blacklist.php
+++ b/src/Blacklist.php
@@ -55,13 +55,12 @@ class Blacklist
     {
         $exp = Utils::timestamp($payload['exp']);
         $refreshExp = Utils::timestamp($payload['iat'])->addMinute($this->refreshTTL);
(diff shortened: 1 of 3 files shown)


tests/BlacklistTest.php
```diff
@@ -28,6 +29,8 @@ class BlacklistTest extends \PHPUnit_Framework_TestCase
 {
     public function setUp()
     {
+        Carbon::setTestNow(Carbon::createFromTimeStampUTC(123));
```
tymondesigns: Cool, didn't know about this :)

tymondesigns:
I will pull this down locally I think, so I can see how things work better

Thanks!

src/Blacklist.php
```diff
@@ -47,17 +54,13 @@ public function __construct(Storage $storage)
...
-            return false;
-        }
+        // find the number of minutes until the expiration date, plus 1 minute to avoid overlap
+        $minutes = $lastExp->diffInMinutes(Utils::now()->subMinute());
 
-        // add a minute to abate potential overlap
-        $minutes = $exp->diffInMinutes(Utils::now()->subMinute());
-
-        $this->storage->add($payload['jti'], ['valid_until' => $this->getGraceTimestamp($exp)], $minutes);
+        $this->storage->add($payload['jti'], ['valid_until' => $this->getGraceTimestamp(Carbon::now())], $minutes);
```
tymondesigns: think it would be good to use `Utils::now()` instead of `Carbon::now()` directly just to keep things consistent

tdhsmith: Yeah for sure, whoops.

tymondesigns:
Could you squash these commits? and I will merge 

Thanks again

tdhsmith:
Sure thing. I'll be on it in a bit!

PR #230, opened 2015-08-31
Cast string claims to int.

mod3st:
iat, exp, nbf should be parsed to json as an integer


tymondesigns:
Already fixed for the upcoming release.

Thanks though

tymondesigns:
See my comment here - https://github.com/tymondesigns/jwt-auth/pull/166

PR #236, opened 2015-09-03
Added RS256 support

alexw23:
Hi there, I had an issue in that when using openssl I could happily set `secret` to be the private key and it would encode the token just fine. However when decoding it wasn't happy using the private key (openssl_verify(): could not get public key from private key).

I was able to place the public key in the `secret` along with the private key and this worked fine, however then it was unable to encode again. Thus it turns out I needed to add another array variable `cert` which allows for an (optional) public key to decode `rs256`. I added some logic to switch between a secret and cert.


tymondesigns:
Can't see why this wouldn't be available for all algorithms, and not just RS256 ?

jchapple:
+1 to see this merged with support for RS512, etc.

lucasmichot:
Which others algo should be available @tymondesigns ?

tymondesigns:
@lucasmichot currently only `HMAC` algorithms are useable, since they can be signed with just a string. This includes `HS256`, `HS384` and `HS512`.

But `RSA` and `ECDSA` algorithms require a public/private key combo, which currently is not setup from a config standpoint. And this PR is a step towards that, but it needs ironing out a bit more.

tdhsmith:
For reference, here's the official JWS algorithm list: https://tools.ietf.org/html/rfc7518#section-3

@tymondesigns Out of curiosity, are you thinking of eventually supporting JWE tokens?

lucasmichot:
I feel IT could be done after next release

tymondesigns:
Yes it is definitely something I want to support, but I do think after next release is best

tymondesigns:
had some spare time and decided to add in this functionality (b7ad60b47cab5b83e4ad8d0474c031df35dc412d) yay :smile:

PR #271, opened 2015-10-01
Fixes typehint based on comment by @mod3st

crynobone:
Made a mistake there since we actually looking for a class that can
resolved to `cache.store` alias. Laravel can resolve with
`Illuminate\Cache\Repository` or `Illuminate\Contracts\Cache\Repository`
while Lumen can only resolve with
`Illuminate\Contracts\Cache\Repository`

https://github.com/laravel/framework/blob/5.1/src/Illuminate/Foundation/Application.php#L1035
https://github.com/laravel/lumen-framework/blob/5.1/src/Application.php#L1659

Signed-off-by: crynobone <crynobone@gmail.com>


tymondesigns:
Good catch @mod3st. Missed that myself

PR #272, opened 2015-10-01
numeric values for exp nbf iat claims in payload

spescina:
These specs suggest to use numeric values for exp-nbf-iat fields in payload.
https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#rfc.section.4.1.1

Some js/node packages relies on these specs and having string fields cause token rejection.
I've type casted the return value of those three claim methods.

### DIFF
--- a/src/PayloadFactory.php
+++ b/src/PayloadFactory.php
@@ -149,7 +149,7 @@ class PayloadFactory
      */
     public function iat()
     {
-        return Utils::now()->format('U');
+        return (int) Utils::now()->format('U');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


spescina:
I've already merged on develop. Let me know if you want a pull request for it.

tymondesigns:
Thanks but this has already been addressed on develop [here](https://github.com/tymondesigns/jwt-auth/blob/develop/src/Factory.php#L158-L181)

spescina:
ahah! nevermind...

PR #306, opened 2015-11-02
Use array_diff to check presence of required claims

Xlator:
Found and fixed an issue causing the validator not to honour the "required_claims" configuration.

### DIFF
--- a/src/Validators/PayloadValidator.php
+++ b/src/Validators/PayloadValidator.php
@@ -62,7 +62,7 @@ class PayloadValidator extends Validator
      */
     protected function validateStructure(array $payload)
     {
-        if (count(array_diff_key($this->requiredClaims, array_keys($payload))) !== 0) {
+        if (count(array_diff($this->requiredClaims, array_keys($payload))) !== 0) {
             throw new TokenInvalidException('JWT payload does not contain the required claims');
         }
 


tymondesigns:
could you elaborate on the exact issue here?

Xlator:
I discovered the problem when I needed to add support for multiple sites
with distinct access tokens. Upon adding the `aud` claim to the required
claims array in the config file, I noticed that payloads without a
specified audience were still getting through. I solved it by overriding
the `validateStructure` method to use `array_diff`. Since `array_keys`
returns a numerical array of the claim types in the payload, and the
required claims array isn't associative either, `array_diff_keys` would
only ensure that the payload has the correct number of claims, without
checking their type. I assumed this was a simple oversight, so I made a PR
to fix it.

Xlator:
I discovered the problem when I needed to add support for multiple sites with distinct access tokens. Upon adding the `aud` claim to the required claims array in the config file, I noticed that payloads without a specified audience were still getting through. I solved it by overriding the `validateStructure` method to use `array_diff`. Since `array_keys` returns a numerical array of the claim types in the payload, and the required claims array isn't associative either, `array_diff_keys` would only ensure that the payload has the correct number of claims, without checking their type. I assumed this was a simple oversight, so I made a PR to fix it.

tymondesigns:
Sorry for the delay on this, but I think your mistaken here.

If you have the following required claims: `['exp', 'iss', 'aud']`

And your payload contained only `['exp' => 123, 'iss' => 'foo']`

Then
```php
$payloadClaims =  array_keys(['exp' => 123, 'iss' => 'foo']); // = ['exp', 'iss']
$difference = array_diff_key(['exp', 'iss', 'aud'], $payloadClaims); // = ['aud']
```

So when a count is checked of the difference it will return 1 (`aud` claim is missing) and throw an exception as expected.

That make sense?

Xlator:
It will still pass with an arbitrary array of claims that is of the same length as the required claims array.
```php
$payloadClaims = array_keys(['foo' => 123, 'bar' => 'baz', 'qux' => true]); // =  ['foo', 'bar','qux']
$difference = array_diff_key(['exp','iss','aud'], $payloadClaims); // = []
```

tymondesigns:
Ah yes, you are right, thanks

Xlator:
Happy to help! This package has been a huge time saver for me.

PR #315, opened 2015-11-18
Standard traits for standard controllers

carduz:


tymondesigns:
Thank you for your efforts here, but I don't think something like this should be included out of the box.

Ultimately, I am aiming to eventually decouple this package from laravel, and this would detract from that goal.

PR #325, opened 2015-12-03
Added remove token method

modder2:
In some situations, for example in unit testing, we need to clear currently saved token.

### DIFF
--- a/src/JWTAuth.php
+++ b/src/JWTAuth.php
@@ -264,6 +264,14 @@ class JWTAuth
     }
 
     /**
+     * Remove the current token.
+     */
+    public function removeToken()
+    {
+        $this->token = null;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


tymondesigns:
Nice addition, however I think I prefer the name `unsetToken()` as I think it is clearer as to what is happening. Would you mind changing that?

modder2:
@tymondesigns As you want. Done. Can you merge this changes to release version as quickly as possible? Or maybe to do commit from master branch?

modder2:
@tymondesigns How often you make new versions for composer?

tymondesigns:
I follow [semver](http://semver.org/).

The `0.6` version and associated docs is very close to being completed. But in the mean time you can get the latest by requiring `0.6.*@dev` in your `composer.json`.

PR #331, opened 2015-12-05
TokenParser should take into account route parameters

andrzejkupczyk:
… because parsing headers and GET parameters is not enough :smirk:

### DIFF
--- a/src/Http/TokenParser.php
+++ b/src/Http/TokenParser.php
@@ -78,13 +78,23 @@ class TokenParser
     }
 
     /**
+     * Try to get the token from the route
+     *
+     * @return false|string
+     */
+    public function fromRoute()
+    {
+        return $this->request->route()->parameter($this->query, false);
+    }
+
+    /**
      * Try to parse the token from the request query string
      *
      * @return false|string
      */
     public function fromQueryString()
     {
-        return $this->request->input($this->query, false);
+        return $this->request->input($this->query, $this->fromRoute());
     }
 
     /**
--- a/tests/Http/TokenParserTest.php
+++ b/tests/Http/TokenParserTest.php
@@ -11,6 +11,7 @@
 
 namespace Tymon\JWTAuth\Test\Http;
 
+use Mockery;
 use Illuminate\Http\Request;
 use Tymon\JWTAuth\Http\TokenParser;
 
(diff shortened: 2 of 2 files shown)


tymondesigns:
Thanks!

crynobone:
Was this tested with Lumen?

```
{
  "message": "Call to a member function parameter() on array",
  "code": 1,
  "status_code": 500,
  "debug": {
    "line": 87,
    "file": "/home/vagrant/katsana/vendor/tymon/jwt-auth/src/Http/TokenParser.php",
    "class": "Symfony\Component\Debug\Exception\FatalErrorException",
    "trace": [
      "#0 /home/vagrant/katsana/vendor/orchestra/lumen/src/Concerns/RegistersExceptionHandlers.php(54): Laravel\Lumen\Application->handleShutdown()",
      "#1 [internal function]: Laravel\Lumen\Application->Laravel\Lumen\Concerns\{closure}()",
      "#2 {main}"
    ]
  }
}
```

andrzejkupczyk:
I didn't have enough time to get this package working with Lumen however route resolver shouldn't return array according to [this](https://github.com/illuminate/http/blob/master/Request.php#L796).
Did you run phpspec (`it_should_return_the_token_from_route` example)?
Did you try #363 fix?

crynobone:
Both solutions doesn't work on Lumen 5.2.

andrzejkupczyk:
There is no Lumen 5.2 release, nice try :wink: 
It seems that your problem is related to JWT-AUTH middleware or [framework itself](https://github.com/laravel/lumen-framework/issues/291). Dunno. As I said, route resolver shouldn't return an array and this is your case.

crynobone:
https://github.com/laravel/lumen-framework/blob/master/composer.json

So meaning jwt-auth doesn't need to prepare for the upcoming release (which going to be release soon). Good job guys!

crynobone:
What phpspec? This project only loads illuminate components, it not testing against full Laravel or Lumen integration. And tested only with phpunit.

tymondesigns:
@crynobone Don't worry, I will be testing everything thoroughly with both Laravel and Lumen prior to the final `0.6` release.

This is still the develop branch afterall :smile:

tymondesigns:
This issue may well be related to https://github.com/laravel/lumen-framework/issues/291.

But I haven't checked that through myself yet.

PR #347, opened 2015-12-14
Add support for Laravel 5.2

crynobone:
* Add support for Laravel 5.1 and Laravel 5.2
* Update travis-ci to include test on both setup.

~~There one failing on laravel 5.1 test at the moment. I not sure how that would relate with the changes.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,25 @@ php:
   - 7.0
   - hhvm
 
+env:
+  global:
+    - laravel='*'
+
 matrix:
+  include:
+    - php: 5.6
+      env: laravel='5.1'
+    - php: 5.6
+      env: laravel='5.2'
   fast_finish: true
   allow_failures:
     - php: 7.0
 
 before_script:
   - travis_retry composer self-update
-  - travis_retry composer install --no-interaction --prefer-source
+  - if [[ $laravel = '*' ]]; then travis_retry composer install --no-interaction --prefer-source; fi
+  - if [[ $laravel = '5.1' ]]; then travis_retry composer require "illuminate/contracts=5.1.*" --prefer-source --no-interaction --dev; fi
+  - if [[ $laravel = '5.2' ]]; then travis_retry composer require "illuminate/contracts=5.2.*" --prefer-source --no-interaction --dev; fi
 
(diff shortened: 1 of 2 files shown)


tymondesigns:
Thanks for this, why is `paragonie/random_compat` required ?

crynobone:
@tymondesigns 

Because `Str::random()` now utilize `random_byte()` which is PHP7 feature, `paragonie/random_compact` provides the polyfill for the function.

https://github.com/illuminate/support/blob/5.2/Str.php#L233

crynobone:
screen shot 2015-12-18 at 7 26 50 pm

This would be the output if I remove `paragonie/random_compat` deps and set `"prefer-stable": false,` with this changes.

irazasyed:
+1 was having issue today when working with Laravel 5.2.

Fed03:
:+1:

jonagoldman:
+1 Specially now that 5.2 is officially released.

LukasMa:
:+1:  Need this merged :)

tymondesigns:
@crynobone I have changed `Str::random()` to `Str::quickRandom()` so shouldn't need to pull in `paragonie/random_compat` anymore. 

Could you remove that & I will merge this in :+1:

tdhsmith:
Looks good for compatibility, though I think a future Laravel 5.2 "embracing" release would use the new authentication driver system, which seems to have been designed specifically for this package's use-case. [The documentation on it even uses JWTs as an example!](https://laravel.com/docs/5.2/authentication#adding-custom-guards)

tymondesigns:
@tdhsmith yes you are bang on there.. I only caught wind of the Auth changes via listening to the podcast - http://www.laravelpodcast.com/episodes/22314-episode-39-updates-on-all-the-things

But ultimately, we will need a release after this one that hooks into the new Auth guards properly.

Also, great to see that example in the docs! :smile:

tymondesigns:
Actually I'll just get this merged now and remove the ref myself Thanks

andrewmclagan:
Any timeline for native 5.2 auth driver support? Would be more then happy to do a PR if one is not already in build...

PR #350, opened 2015-12-14
Added Auth.php - implementation of Guard

JaffParker:
Created an implementation of Illuminate\Contracts\Auth\Guard for seamless replacement of standard Laravel Auth with JWTAuth.

### DIFF
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -1,4 +1,4 @@
-<?php namespace Koodzo;
+<?php namespace Tymon\JWTAuth;
 
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Contracts\Auth\Guard;


crynobone:
Would actually love to see a new Auth "guard" (driver) based on 5.2 new structure.

JaffParker:
@crynobone I'll check it out

jonagoldman:
+1 Specially now that 5.2 is officially released.

src/Auth.php
```diff
@@ -0,0 +1,171 @@
+<?php namespace Tymon\JWTAuth;
+
+use Illuminate\Contracts\Auth\Authenticatable;
+use Illuminate\Contracts\Auth\Guard;
+use Illuminate\Http\Request;
+use Koodzo\Models\Db\User;
```
crynobone: I don't think this is a generic `User` model.

tymondesigns:
closing as I will be looking to add proper support for the Auth Guard stuff that was added in 5.2

PR #363, opened 2015-12-29
Fixed parsing token from the route

andrzejkupczyk:
@VeerababuKottu pointed out that parsing token from the route in that way could throw an exception.
This fix should help prevent it from happening.


tymondesigns:
Things changed c3dea4fdcae42ff7c5cf16966edbb04a3792e4ac and have added this in myself

Thanks!

andrzejkupczyk:
Great, thank you.

PR #373, opened 2016-01-07
Switch to new recommended composer syntax.

vinkla:
Composer now recommends to use `||` instead of just one `|`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,17 +20,17 @@
     ],
     "require": {
         "php": ">=5.5.9",
-        "illuminate/support": "5.1.*|5.2.*",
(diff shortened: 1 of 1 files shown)


tymondesigns:
Thanks!

vinkla:
:+1:

vinkla:
Btw, are you planning to integrate the new auth features in 5.2?

tymondesigns:
that's the plan, probs gonna get to next week before I can dig in

vinkla:
Okay, good. We're also planning it for the OAuth2 bridge.

PR #380, opened 2016-01-08
Provide support for tagged caches and squash L5.1.28 breakage

### LINKED ISSUE #367: In 0.6.*@dev + L5.1, there is a problem about cache tag.
An error occurs.
"message": "This cache store does not support tagging."
The problem was found, in vendor/tymon/jwt-auth/src/Providers/Storage/Illuminate.php:103, the cache() function.

```php
    protected function cache()
    {
        if (! method_exists($this->cache, 'tags')) {
            return $this->cache;
        }

        return $this->cache->tags($this->tag);
    }
 ```
The $this->cache is \Illuminate\Cache\Repository and it always has a tags() function. The method_exists() should be applied to $this->cache->getStore(). But after the modification, the test case failed because the getStore() is not in the CacheContract.

tdhsmith:
Existing checks for whether a cache supported `tags()` were quietly failing because it was handled by magical `__call` in L5.1.27 and lower. Instead of a raw method exists, this PR captures exceptions on `tags` (to handle L5.1.28) and also examine the store if a `getStore` method is available (for older releases). This fixes #367.

I also uncovered why this hadn't been caught by the tests: Using the "Demeter chain" syntax wasn't actually checking whether `tags()` was getting called, because Mockery shortcuts those checks when it knows what the output would be and only verifies the final method in the chain. Instead we separate the chain method expectations, as well as enforcing they are actually called (with `once()`) and duplicating them to have separate tests for the tagged and untagged formats.

### DIFF
--- a/src/Providers/Storage/Illuminate.php
+++ b/src/Providers/Storage/Illuminate.php
@@ -25,6 +25,11 @@ class Illuminate implements Storage
      * @var string
      */
     protected $tag = 'tymon.jwt';
+    
+    /**
+     * @var boolean
+     */
+    protected $supportsTags;
 
     /**
      * @param \Illuminate\Contracts\Cache\Repository  $cache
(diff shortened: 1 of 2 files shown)


tymondesigns:
Thanks for your time on this buddy :+1:

Issue #384, opened 2016-01-09
Lumen 5.2 Issues!


irazasyed:
So I've been testing this package in Lumen 5.2 and i came across the following issues:

**1. Class cache.store does not exist**
**Path:** `Tymon\JWTAuth\Providers\Storage\Illuminate`
> So the issue with this is that, with `Illuminate\Contracts\Cache\Repository as CacheContract` in constructor, it fails for some reason but if that is replaced with simple `app('cache')`, it works fine. Weird and not sure what's happening.

**2. Fatal error: Call to undefined method Illuminate\Auth\RequestGuard::onceUsingId()**
**Path:** `Tymon\JWTAuth\Providers\Auth\Illuminate`

Is anyone else facing these issues? Is there a solution?

tdhsmith:
> So the issue with this is that, with Illuminate\Contracts\Cache\Repository as CacheContract in constructor, it fails for some reason but if that is replaced with simple app('cache'), it works fine. Weird and not sure what's happening.

Hrm I'm not sure what's up there; I know very little about Lumen but they both should be using the same Cache module. I guess it could just be a difference in bindings/DI? In any case, you should probably be using `app('cache.store')` because `cache` alone points to the cache _factory_ interface, and I'm pretty certain that's going to break once you rely on it.

> 2\. Fatal error: Call to undefined method Illuminate\Auth\RequestGuard::onceUsingId()
Path: Tymon\JWTAuth\Providers\Auth\Illuminate

Urgh ok this looks a lot trickier. Lumen "doesn't support" session-based auth, so by default it has you extend Auth with a custom, callback-based driver (see [the first section in the docs here](https://lumen.laravel.com/docs/5.2/authentication)). Unfortunately, the `Guard` for this driver, `RequestGuard`, doesn't implement the methods we need. We don't actually care about sessions or statefulness (JWT is stateless), but we use `Auth::onceUsingId($credentials)` to support our User lookups (`JWTAuth::authenticate`).

So for now you either can not use `authenticate` and do your user lookups manually from the `sub` claim, or you can write a custom provider implementing `Tymon\JWTAuth\Contracts\Providers\Auth` that doesn't utilize `onceUsingId` to find the user.

tdhsmith:
I think the long-term solutions to the second problem are:

1. Native support as an 5.2 auth driver
[...]

Issue #376, opened 2016-01-08
Add support & documentation for integrating with Lumen 5.2


rajabishek:
Since Lumen does not support session state, incoming requests that we wish to authenticate must be authenticated via a stateless mechanism such as API tokens. This package is really going to be helpful for resolving out the authenticated user from the request through API token.
https://lumen.laravel.com/docs/5.2/authentication

``` php
Auth::viaRequest('api', function ($request) {
    try {
        if (! $user = JWTAuth::parseToken()->authenticate()) {
            return null;
        }
    } catch (Exception $e) {
        return null;
    } 
    return $user;
});
```

It would be really helpful if we have another section in the wiki about integrating this package with Lumen. I see that most of the work has already been done ( already there is a LumenServiceProvider ), only that it needs to be upgraded to support v5.2 and documented. Thank you Symon for giving us this wonderful package that we can use with out Laravel and Lumen projects.

syropian:
@rajabishek Where do we put this `Auth::viaRequest` function? And what exactly is `api` here?

syropian:
Weird, $user always return `null ` for me. It's the `null` from the `if` statement, not the `catch` though.

rajabishek:
@syropian The AuthServiceProvider in Lumen 5.2 already has a call to Auth::viaRequest method that accepts a Closure which will be called when the incoming request needs to be authenticated ( if auth middleware is used ). Within this Closure, we may resolve out App\User instance however we wish. If no authenticated user can be found for the request, the Closure should return. So what I was suggesting was if we use this package with Lumen 5.2 then we would have to do something along the lines of 
```php
Auth::viaRequest('api', function ($request) {
    try {
        if (! $user = JWTAuth::parseToken()->authenticate()) {
            return null;
        }
    } catch (Exception $e) {
        return null;
    } 
    return $user;
});
```
And now once this is done, in our controllers or route handling closures we can directly get the authenticated user using the ```Auth::user()``` or through ```$request->user()``` just like how we would deal with sessions. Something like
```php
use Illuminate\Http\Request;

$app->get('/post/{id}', ['middleware' => 'auth', function (Request $request, $id) {
    $user = Auth::user();
    //or
    $user = $request->user();
}]);
```

So I just felt there must be a separate section in the wiki that needs to have this documented on how we can integrate and use this package with Lumen apps.

syropian:
[...]

PR #391, opened 2016-01-13
Make route parameter parsing safer

### LINKED ISSUE #361: Call to a member function parameter() on null
I use development branch to use package with doctrine and its private properties.
Since last update I get this error:
~~~
Call to a member function parameter() on null in /Project/vendor/tymon/jwt-auth/src/Http/TokenParser.php on line 87
~~~

This is the code that trown this error
~~~
        $httpRequest = new Request( [
            'token' => JWTAuth::fromUser($this->user)
        ]);

        \JWTAuth::parser()->setRequest($httpRequest)->hasToken();
        \JWTAuth::parseToken()->authenticate();
~~~

tdhsmith:
Added `is_callable` check for `parameter` method that was throwing fatal errors when trying to parse Lumen requests and requests that had never been dispatched, along with appropriate tests.

Also allow parsing chain to be retrieved using `JWTAuth::parser()->getChain()`, and provide a small experimental class to support unstable Lumen route params.

Addresses components of #384 & #376, and fixes #361.

### DIFF
--- a/src/Http/LumenRouteParams.php
+++ b/src/Http/LumenRouteParams.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of jwt-auth
+ *
+ * (c) Sean Tymon <tymon148@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Tymon\JWTAuth\Http;
+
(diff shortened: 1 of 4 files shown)


tymondesigns:
Thanks, hopefully the Lumen issue will be resolved soon, then we can remove `LumenRouteParams`

Issue #397, opened 2016-01-18
everytime user_not_found


ilvalerione:
The routes after jwt middleware return everytime "user_not_found".

1) Primary key of my User object is "id" how to standard jwt configiguration.

2) I tried to send token via HTTP Header and into GET request but i get always the same response => "user_not_found"

3) When user logged-in I create a token with "fromUser" method

Issue #393, opened 2016-01-15
error: "user_not_found" on multiple async api calls


MateEke:
Hi!

I have the same problem as in this stackoverflow question:
http://stackoverflow.com/questions/34451540/jwt-tymon-laravel-5-1-angularjs-multiple-async-call-error-user-not-found

The sent authorization header is the same in all of the requests, but some of them returns {"error":"user_not_found"}.

PR #398, opened 2016-01-18
Bring auth-exception-bubbling fix to 0.5

tdhsmith:
I'm concerned that this try-catch that got removed on develop (6f7ccb7d0c28da416831fcaac27e0a4e5d242bee) is impeding debugging issues people are having on master as well, because Auth exceptions show up through the middleware there as `user_not_found`. This could be the root cause of #393 & #397 for example...

### DIFF
--- a/src/Providers/Auth/IlluminateAuthAdapter.php
+++ b/src/Providers/Auth/IlluminateAuthAdapter.php
@@ -2,7 +2,6 @@
 
 namespace Tymon\JWTAuth\Providers\Auth;
 
-use Exception;
 use Illuminate\Auth\AuthManager;
 
 class IlluminateAuthAdapter implements AuthInterface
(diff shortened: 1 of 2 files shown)


tymondesigns:
Good idea! but looks like tests need updating [here](https://github.com/tymondesigns/jwt-auth/blob/master/tests/Providers/Auth/IlluminateAuthAdapterTest.php#L35-L40)

tdhsmith:
Derp that's what I get for throwing up an edit without testing. Bad form, Taylor.

I'll get a test up when I get home in a few hours.

tymondesigns:
s'all good :+1: I do it too

tdhsmith:
OK passing tests now, and there's a distinct case for bubbling.

tymondesigns:
Cheers!

Issue #402, opened 2016-01-22
Laravel PHPUnit Error Event Dispatacher


kontrollanten:
When I use expectsEvents function i Laravel PHPUnit the jwt auth middleware throws an error.

This is the test:

```
    public function testPutCarts()
    {
        $cart = factory(App\Cart::class)->create();
        $this->expectsEvents(App\Providers\Events\CartStatusWasUpdated::class);
        $response = $this->call(
            'PUT', 
            $this->endpoint . '/' . $cart->id, 
            ['status' => $status],
            [],
            [],
            $headers);
    }
```
 And this is the response:

```
ErrorException: Argument 2 passed to Tymon\JWTAuth\Middleware\BaseMiddleware::__construct() must be an instance of Illuminate\Events\Dispatcher, instance of Mockery_0_Illuminate_Contracts_Events_Dispatcher given

vendor/tymon/jwt-auth/src/Middleware/BaseMiddleware.php:33
```

This issue seems like a similar https://github.com/cartalyst/sentinel/issues/166 which they solved in this commit https://github.com/cartalyst/sentinel/commit/3f2e7e89c5f85bd755b38875f6810bcfb0bf015c

tdhsmith:
Ahh yes it looks like `expectsEvents` [mocks the contract](Illuminate\Contracts\Events\Dispatcher) and we're [requiring the implementer](https://github.com/tymondesigns/jwt-auth/blob/master/src/Middleware/BaseMiddleware.php#L33). I'll double-check if there's any important discrepancies or side effects between the two later today and see if I can get a simpler test case going.

kontrollanten:
I couldn't contribute to the repo so I've attached a diff that uses the contracts instead. It solves my problem and it works with the phpunit tests :-) 

[#402.diff](https://github.com/tymondesigns/jwt-auth/files/101628/402.txt)

tymondesigns:
merged and bundled with `0.5.7`

PR #405, opened 2016-01-25
Use contracts for dispatch and response in base middleware

tdhsmith:
This fixes errors with class mismatch of mocked components as described in #402. There it was triggered by `$laravelTestCase->withoutEvents()` but I think there are other methods that might've had problems.

Also addresses the extremely unlikely but still valid issue of compatibility for anyone using a non-stock implementation of `Events\Dispatcher` or `Routing\ResponseFactory`. 

(Unfortunately building a test specifically for this would require including all of Laravel, because, as far as I can tell, there isn't a standalone repo that contains `Illuminate\Foundation\Testing`.)

Thanks to @kontrollanten for the diff.

### DIFF
--- a/src/Middleware/BaseMiddleware.php
+++ b/src/Middleware/BaseMiddleware.php
@@ -3,18 +3,18 @@
 namespace Tymon\JWTAuth\Middleware;
 
 use Tymon\JWTAuth\JWTAuth;
-use Illuminate\Events\Dispatcher;
-use Illuminate\Routing\ResponseFactory;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Contracts\Routing\ResponseFactory;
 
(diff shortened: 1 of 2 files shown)


tdhsmith:
I'll have a PR with the latter half of the change onto develop in a bit. (Since there are no events there, the former doesn't matter anymore.) I wanna look around for other contract issues like this first though.

tdhsmith:
I'm not thinking right about code today: develop doesn't use `ResourceFactory` either. So none of this really applies to 0.6.

I still checked around for Illuminate components that might be replaceable with their interfaces, but pretty much everything already has the contract version or is a core structure without an interface (`Request`, `Collection`, `Arr`, etc).

tymondesigns:
No worries. Thanks for that

PR #407, opened 2016-01-26
Throws `JWTException` type on required token absent

jfoliveira:
Current implementation is throwing a `BadRequestHttpException`.

Although I agree this is a fault on the client side that is not including the required access token when submitting the request, having `JWTGuard` throwing a package specific exception type makes it easier to intercept the exception and customize its message as needed.

Also that keeps the implementation consistent with the [examples given on the wiki](https://github.com/tymondesigns/jwt-auth/wiki/Authentication#retreiving-the-authenticated-user-from-a-token).


tymondesigns:
Yea i was actually gonna change this myself, so thanks for saving me a job 😉

PR #411, opened 2016-01-26
Use a command signature

lucasmichot:
Since L5.1, we can use command signatures which are much more readable.


tymondesigns:
Thanks, didn't know about this one :+1:

PR #413, opened 2016-01-26
Cleanup composer.json

lucasmichot:
* Sort keywords, require and require-dev
* Suffix folders

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,12 +36,12 @@
     },
     "autoload": {
         "psr-4": {
-            "Tymon\\JWTAuth\\": "src"
+            "Tymon\\JWTAuth\\": "src/"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "Tymon\\JWTAuth\\Test\\": "tests"
+            "Tymon\\JWTAuth\\Test\\": "tests/"
         }
     },
     "extra": {
--- a/composer.json
+++ b/composer.json
@@ -2,11 +2,11 @@
     "name": "tymon/jwt-auth",
     "description": "JSON Web Token Authentication for Laravel and Lumen",
     "keywords": [
-        "jwt",
         "auth",
         "authentication",
-        "laravel",
-        "json web token"
+        "json web token",
+        "jwt",
+        "laravel"
     ],
     "homepage": "https://github.com/tymondesigns/jwt-auth",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1:

lucasmichot:
Comment?

tymondesigns:
@vinkla added the double pipes in #373 see [here](https://getcomposer.org/doc/articles/versions.md#range)

What's the difference between `5.1.* || 5.2.*` and `~5.1.0|~5.2.0` ?

composer.json
```diff
@@ -20,28 +20,28 @@
     ],
     "require": {
         "php": ">=5.5.9",
-        "illuminate/support": "5.1.* || 5.2.*",
-        "illuminate/http": "5.1.* || 5.2.*",
-        "illuminate/auth": "5.1.* || 5.2.*",
-        "illuminate/contracts": "5.1.* || 5.2.*",
-        "namshi/jose": "6.0.*",
-        "nesbot/carbon": "^1.0"
+        "illuminate/auth": "~5.1.0|~5.2.0",
```
vinkla: This is dangerous since we don't know if it will break in future versions of Laravel.

vinkla:
Yeah, I'm also :-1: on this.

vinkla:
`5.1.* || 5.2.*` will allow only those specific versions of Laravel. `||` is now recommended by Composer instead of one `|` as we did before.

composer.json
lucasmichot: How can this be dangerous ? We will not reach `5.3.*`

crynobone: It's not going to install 5.3 when it's available with or without the changes.

vinkla:
I agree we can suffix the folder paths :+1: .

composer.json
lucasmichot: @vinkla see: http://blog.madewithlove.be/post/tilde-and-caret-constraints/

vinkla: Yeah, I read it wrong. Sorry, removed the comment.

tymondesigns:
So I guess we can leave the version declarations as is, since there is no difference either way: 

5.1._whatever_ OR 5.2._whatever_

happy with everything else

lucasmichot:
@vinkla the `||` notation seems specific to the PHP version, see https://getcomposer.org/doc/04-schema.md#require
I cannot find any assertion of what you say about the `||` notation either in the composer docs, either in https://github.com/composer/composer issues or PRs

The only thing that I see is that a support is offered to npm-like notation in an alpha release : https://github.com/composer/composer/releases/tag/1.0.0-alpha10 

Do you have more infos ?

GrahamCampbell:
:-1:

GrahamCampbell:
If it's not broke, don't change it.

vinkla:
> Yeah `||` is the new cool, npm compatible and the only thing we recommend in docs https://getcomposer.org/doc/articles/versions.md#range
https://github.com/composer/semver/pull/22#issuecomment-151858703

lucasmichot:
> If it's not broke, don't change it.

@GrahamCampbell 

* Sorting keywords and require allow to avoid noise when adding or removing keywords and/or requires
* adding the / suffix allows composer not to do it

Also making things more readable, organized or usage-consitant are not useless

PR #415, opened 2016-01-26
Also check /tests files

lucasmichot:
Also check /tests files


lucasmichot:
FYI, styleci also checks tests/

tymondesigns:
good catch, Thanks!

PR #427, opened 2016-01-27
No need to export .editorconfig

lucasmichot:
No need to export `.editorconfig`


lucasmichot:
Travis CI seems to fail for another reason.

PR #430, opened 2016-01-27
Use @expectedException notation

lucasmichot:
Use `@expectedException` notation


tymondesigns:
I guess this is slightly cleaner, thanks

PR #431, opened 2016-01-27
Remove unused variables in tests

lucasmichot:
Remove unused variables in tests


tests/Providers/JWT/NamshiTest.php
```diff
@@ -46,7 +46,7 @@ public function it_should_return_the_token_when_passing_a_valid_subject_to_encod
         $this->jws->shouldReceive('sign')->once()->with('secret')->andReturn(Mockery::self());
         $this->jws->shouldReceive('getTokenString')->once()->andReturn('foo.bar.baz');
 
-        $token = $this->provider->encode($payload);
```
tymondesigns: this var is needed

PR #438, opened 2016-01-27
No need to use array_has

lucasmichot:
Directly use \Illuminate\Support\Arr::has


tymondesigns:
I'll save you resolving this conflict :smile: 

And thank you for the wave of PR's that you've sent over the last couple of days :+1: :boom:

PR #439, opened 2016-01-27
Sync .php_cs

lucasmichot:
@tymondesigns Apply the same settings than styleci.
So that the developer can fix code before pushing, and styleci only fix when this is necessary.


tymondesigns:
Saved me a job :+1:

lucasmichot:
You're welcome !

GrahamCampbell:
This isn't quite synced since laravel's .php_cs file is out of date, and styleci no longer matches up with php-cs-fixer tbh.

tymondesigns:
What would you suggest Graham? Is there a way to sync styleci settings with a locally tool?

GrahamCampbell:
Not really. I'd recommend just ditching any local tools and turning on auto-fixing and merging on StyleCI. That way it'll just straight up merge fixes for you automatically.

GrahamCampbell:
That was really the entire point of StyleCI. Not to compliment local tools, but to totally remove them.

tymondesigns:
Yea i had a feeling that would be the answer :smile: 

And its a perfectly valid point, I'll think on it. Thanks

GrahamCampbell:
One advantage for example is someone can worry free make changes even from github's online editor, and say, if they left some whitespace in, it'll be fixed up in seconds automatically.

PR #440, opened 2016-01-27
Consistant comparison

lucasmichot:
Remains consistant with the syntax used for `\Tymon\JWTAuth\Validators\PayloadValidator::validateTimestamps`


tymondesigns:
:+1:

Issue #378, opened 2016-01-08
How to use refresh token (RefreshToken Middleware)


bravist:
routes.php

``` php
$api->group(['before' => 'jwt.auth', 'after' => 'jwt.refresh'], function($api) {

                $api->get('users', 'AuthenticateController@index');
                $api->get('users/{id}', 'AuthenticateController@show');

            });
```

jwt.php

``` php

'ttl' => 1,

'refresh_ttl' => 3,
```

After 1 minute, token does not expired, but after 3 minutes, token can also be used.Don't know how to solve, please help to explain why

imrealashu:
just use middleware `jwt.auth`. it will work well

tdhsmith:
@Brave-Cheng Sorry for missing this issue. Are you saying 3 minutes after a refresh, or 3 minutes from the initial token creation? I made a diagram [here](https://github.com/tymondesigns/jwt-auth/issues/201#issuecomment-131978407) about how refreshes are supposed to work in relation to `ttl` and `refresh_ttl`.

Walk-Code:
**You can try to build an authJWT middleware and register it like this:**
public function handle($request, Closure $next)
    {
        try {
            $user = JWTAuth::toUser($request->input('token'));
        } catch (Exception $e) {
            if ($e instanceof \Tymon\JWTAuth\Exceptions\TokenInvalidException){
                return response()->json(['error'=>'token_invalid']);
            }else if ($e instanceof \Tymon\JWTAuth\Exceptions\TokenExpiredException){

                $token = JWTAuth::getToken();
                if (!$token) {
                    return $this->response->errorMethodNotAllowed('Token not provided');
                }
                try {
                    $refreshedToken = JWTAuth::refresh($token);
                } catch (JWTException $e) {
                    return $this->response->errorInternal('Not able to refresh Token');
                }

                return response()->json(['error'=>'token_expired','token'=>$refreshedToken]);
            }else{
                return response()->json(['error'=>'error']);
            }
        }
        return $next($request);
    }

PR #441, opened 2016-01-27
Allow expired-but-refreshable tokens to be blacklisted (again)

### LINKED ISSUE #419: Unlimited refreshes during refresh_ttl ?
Hi,
I'm beginning with jwt and I wonder if this behaviour is normal : 

When your token is expired (> ttl) but still in refresh window (< refresh_ttl) you are able to refresh it as many times as you want. Nothing blocks you, as only unexpired tokens are put in blacklist when refreshing.

Shouldn't you be able to refresh it only once ?

tdhsmith:
Remake of the fix way back in #218 that I neglected to realize only ever made it to develop and not master. :dizzy_face:

Now expired-but-refreshable tokens can be blacklisted, which should fix a few refreshes people are seeing that should not succeed! Fixes #419, potentially related to some confusing situations like #378.

### DIFF
--- a/src/Blacklist.php
+++ b/src/Blacklist.php
@@ -19,6 +19,13 @@ class Blacklist
      * @var \Tymon\JWTAuth\Providers\Storage\StorageInterface
      */
     protected $storage;
+    
+    /**
+     * Number of minutes from issue date in which a JWT can be refreshed.
+     *
+     * @var int
+     */
+    protected $refreshTTL = 20160;
 
     /**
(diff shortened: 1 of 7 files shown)


tymondesigns:
Awesome, thanks!

I guess I should release a patch version with these recent fixes!

tdhsmith:
Yeah this and #398, among others, might merit a 0.5.7, at least to make debugging issues easier until 0.6 is stable. :+1:

PR #445, opened 2016-01-27
Stay consistant with constructor without arguments

lucasmichot:
Stay consistant with constructor without arguments

### DIFF
--- a/src/Providers/LumenServiceProvider.php
+++ b/src/Providers/LumenServiceProvider.php
@@ -234,7 +234,7 @@ class LumenServiceProvider extends ServiceProvider
     protected function registerJWTCommand()
     {
         $this->app->singleton('tymon.jwt.secret', function () {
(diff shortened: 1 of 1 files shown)


tymondesigns:
I really need to fix these tests randomly failing because of timings

lucasmichot:
@tymondesigns you should use `Carbon` with setTestNow
I could solve these kind of issues with that

tymondesigns:
Yea I've been thinking about doing it for a while.. but not got round to it

tymondesigns:
it always seemed like it would be a pain to do.. probably not the case though

lucasmichot:
I'll have a try. Anyway I still have to make a heavy use of this package for the next days.
If i can solve that, I'll submit a PR

tymondesigns:
Cool, appreciate it

lucasmichot:
Good night ;-)

PR #459, opened 2016-01-29
Updated concatenation to match laravel/symfony

GrahamCampbell:


GrahamCampbell:
See the StyleCI diff for what this will look like.

tymondesigns:
yea why not, I'm averse to this, and it's one less exception

PR #460, opened 2016-01-29
[WIP] Fix Travis build no. 105635086

lucasmichot:
Fix failing tests with L5.1 https://travis-ci.org/tymondesigns/jwt-auth/jobs/105635086
Note this can be a BC, as `\Tymon\JWTAuth\JWTAuth` could be extended.

Strangely last 5.1 release is https://github.com/laravel/framework/releases/tag/v5.1.27 on December 17, And attempt was not modified since https://github.com/tymondesigns/jwt-auth/commit/20d1ee4568a79aaaaf5132c471e5cc177eea4385

Trying to figure out why the test was green before

### DIFF
--- a/src/JWTAuth.php
+++ b/src/JWTAuth.php
@@ -35,11 +35,12 @@ class JWTAuth extends JWT
     /**
      * Attempt to authenticate the user and return the token.
      *
-     * @param array $credentials
-     *
+     * @param  array  $credentials
+     * @param  bool   $remember
+     * @param  bool   $login
      * @return false|string
      */
-    public function attempt(array $credentials)
+    public function attempt(array $credentials = [], $remember = false, $login = true)
     {
         if (! $this->auth->byCredentials($credentials)) {
             return false;


tymondesigns:
The reason the test was green is because I hadn't added any tests for JWTGuard until recently.

So obviously JWTGuard is not intended to be used with laravel 5.1.

I'll need to think of a new strategy...

lucasmichot:
Hmmm... make sense

tymondesigns:
One way would be to pluck out the JWTGuard to its own package, but I don't think i'll be doing that

lucasmichot:
Tricky choice, I would not do so either.
I tend to think that making separate branches is always easier to manage

jwt 1.1 -> L5.1
jwt 1.2 -> L5.2

But it can also have disadvantages

tymondesigns:
Closing for now, since It makes sense to use the latest contract (5.2) here.

Still not decided on how i'll handle this issue, but if I do go down the route of separating by version, then I think it would be necessary to pull out a "core" piece of the package to save trying to maintain 2 variants

PR #462, opened 2016-01-29
Raises `JWTException` on BaseMiddleware

jfoliveira:
Part 2 of #407 `Throws "JWTException" type on required token absent `


src/Middleware/BaseMiddleware.php
```diff
@@ -37,12 +37,12 @@ public function __construct(JWTAuth $auth)
...
      * @param   \Illuminate\Http\Request $request
      *
-     * @throws \Symfony\Component\HttpKernel\Exception\BadRequestHttpException
+     * @throws \Tymon\JWTAuth\Exceptions\JWTException
      */
     public function checkForToken(Request $request)
     {
         if (! $this->auth->parser()->setRequest($request)->hasToken()) {
-            throw new BadRequestHttpException('Token not provided');
+            throw new JWTException('Token not provided');
```
tymondesigns: I don't think this is the same as #407, because this is within middleware and a http specific exception is relevant here imo

src/JWTAuth.php
```diff
@@ -57,11 +57,13 @@ public function authenticate()
     {
         $id = $this->getPayload()->get('sub');
 
-        if (! $this->auth->byId($id)) {
-            return false;
+        $retrieveUser = config("jwt.retrieve_user_on_authentication", true);
```
tymondesigns: It might sound strange, given that this package is aimed at LaraveI, but I don't really want to bake in the config function here.

You'll see that i've taken care not to do things like this throughout the package, and i've delegated all of the config to the service providers where it belongs :)

tymondesigns: This would be better placed as a parameter I think.

```php
public function authenticate($retrieveUser = true)
{}
```

src/Middleware/BaseMiddleware.php
jfoliveira: @tymondesigns I thought of that, that middle-wares should thrown HTTP exception. I'm trying to find a way to allow the application to intercept/localize the message though.
It would be a better approach to pass an exception code as the third argument to `BadRequestHttpException`?
Or add configuration to customize the message?

Or is up to the app code to detect this specific exception by comparing the message string:
```php
if ($e instanceof BadRequestHttpException && $e->getMessage() == 'Token not provideed') {
...
}
```
?

tymondesigns: I would recommend simply adding your own middleware if you require different behaviour tbh. Middleware is a very application specific thing afterall.

Issue #451, opened 2016-01-28
Non expiring token as API Key - how manage multiple guard settings?


jfoliveira:
I'm using latest `0.6.*@dev` with `Lumen 5.2` and I trying to figure out how could I support two token types on my API:

1) User token: that will be retrieved every time an user "logs into the API" and will be blacklisted when the users logs out.

2) API Key: that should be issued once and given to a third party. This one should never expire as a regular API Key.

**Question:**
If I remove `exp` claim from my `required_claims` the tokens are issued without expiration date (API Key).
Keeping this claim the tokens are issued with `exp` claim (user token).

**How to set/remove the `exp` claim programmatically** or manage different settings for each `Auth::guard`?

Below some code that illustrates my current scenario that is working for user tokens only.

`config/auth.php`:

``` php
<?php

return [
    'defaults' => [
        'guard' => env('AUTH_GUARD', 'user_token'),
        'provider' => 'users',
    ],
    'guards' => [
        'user_token' => [
            'driver' => 'jwt',
            'provider' => 'users',
        ],
        'api_key' => [
            'driver' => 'jwt',
            'provider' => 'consumers',
        ],
    ],

    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\Models\User::class,
        ],
        'consumers' => [
            'driver' => 'eloquent',
            'model' => App\Models\Consumer::class,
        ],
    ],
];
```

`config/jwt.php`:

``` php
<?php

return [
    'secret' => env('JWT_SECRET'),
    'ttl' => env('JWT_TTL', 60),
    'refresh_ttl' => env('JWT_REFRESH_TTL', 20160),
    'algo' => env('JWT_ALGO', 'HS256'),
    'required_claims' => ['iss', 'iat', 'exp', 'nbf', 'sub', 'jti'],
    'blacklist_enabled' => env('JWT_BLACKLIST_ENABLED', true),
    'blacklist_grace_period' => env('JWT_BLACKLIST_GRACE_PERIOD', 10),
    'providers' => [
        'jwt' => Tymon\JWTAuth\Providers\JWT\Namshi::class,
        'auth' => Tymon\JWTAuth\Providers\Auth\Illuminate::class,
        'storage' => Tymon\JWTAuth\Providers\Storage\Illuminate::class,
    ],
];
```

`My token generator`:

``` php
public static function getToken($subjectId, $tokenType)
{
    $token = null;

    switch ($tokenType) {
[...]

Issue #275, opened 2015-10-02
Avoid query DB


jepezi:
This is not really an issue but a question about strategy for protected route. I'm using Laravel for api server and have a react app on the front on my personal project for learning purpose. I use the `GetUserFromToken` middleware for protected routes. Everything works great. From my understanding, every time auth user hits the protected route, the middleware finds token in the request and queries DB for user based on `sub` (id in this case), then let it pass if found one. 

Now I'm wondering if it's a good idea to, for example, create the middleware to just decode (verify) the token and grab the user encoded in the token, instead of hitting DB when I'm already using token. In other word, why would we query DB for user when we already have everything we need in token (for protected routes that is)?

Sorry if I've missed something very obvious. Thanks!

Deleugpn:
I don't know if I got it right and (also) my experience with e rest api is limited, but being stateless makes your API require to make sure that everything is in order. From my understanding, each request is a new unrelated communication. It's like the server doesn't even know what's your token is about. I see it like those places you buy a storage box, people there give you a key and you go away. When you come back, they don't know who you are or where you got the key, so they test it in the keyhole and if it works you have the right for the content of the storage.
When the api checks the database to make sure the user exists, it's doing the keyhole job.

jepezi:
Thanks for your opinion. To further discuss, let's continue with your keyhole example but with my slightly different view. 

What if the keyhole job is the token verification process? i.e. Key (token) is considered invalid when the verification fails (key is expired, signature is invalid, ..) In other words, I think keyhole job is about verifying the token, not the auth attempt. You have the valid key, you get the content decoded in the key (user info you put in the first place) and you go ahead.

harshitdkanodia:
We can use custom attributes set the user info in the token itself , but its not good practice !
Say for example  you have a field in your database that indicates whether the user has certain rights .
Let's say your user logged in and his information is in the token, now he's using it and someone changes his rights in the database ..
Now the token and database are not in sync .
In my opinion it's a good practice to only include information that is absolutely necessary in the token , you would not put all the use info for the same reason you would set a foreign key on a column 
To keep everything in sync and validated .

*peace*

jfoliveira:
I'm new to JWT and as @jepezi I was also under the impression that querying the DB on every request could be avoided.

Quoting http://jwt.io/introduction/:

>**Self-contained**: The payload contains all the required information about the user, **to avoid querying the database more than once**.
> ...
[...]

PR #464, opened 2016-01-29
Configuration to retrieve user on authentication or avoid querying DB

jfoliveira:
As discussed in  #275 `Avoid query DB` and #451`Non expiring token as API Key - how manage multiple guard settings?` it seems there are mixed feelings about which approach would be the ideal about retrieving or not the user on authentication.

This is my proposal to allow this behavior to be customizable on the configuration file.

Please review it. Feedback is more than welcomed.


jfoliveira:
@tymondesigns Regarding your comments on #462 referring to this branch's changes (my bad, my original commit to 462 was messed up):

1. I refactored the code, reading the configuration from the service provider. Thanks for pointing that out!
2. Regarding pass the `$retrieveUser` flag as a parameter to the `authenticate()` method, it seems to me that could lead to inconsistency as each caller should decide whether to read the config key or not to determine the parameter value - which doesn't sound like a good idea to me. What do you think?

src/Manager.php
```diff
@@ -44,6 +44,11 @@ class Manager
     protected $blacklistEnabled = true;
 
     /**
+     * @var bool
+     */
+    protected $retrieveUserOnAuthentication = true;
```
tymondesigns: I don't think this should be on the `Manager` class, as it is authentication specific, and the Manager has no knowledge or dealings with auth mechanisms.

Instead it should go on the `JWTAuth` class

config/config.php
```diff
@@ -112,6 +112,25 @@
...
+    | existing user the user could be automatically retrieved from the database
+    | every time a token is validated.
+    |
+    | Setting it to "true" will retrieve the user every time the token is checked.
+    | Setting it to "false will query the database only when Auth::user() is invoked
+    | for the first time.
+    |
+    */
+
+    'retrieve_user_on_authentication' => true,
```
tymondesigns: I'm afraid |'m still not sold on adding this to the global config, as it is quite a unique requirement for most people I would say. And I don't want to add too much in here, as there are already lots of options.

I would class it more as advanced config and that has it's place in a custom service provider in my eyes.

Something like:

```php
// In your AppServiceProvider or something
app('tymon.jwt.auth')->setRetrieveUserOnAuthentication(true);
```

Thoughts?

src/JWTAuth.php
```diff
@@ -59,11 +59,13 @@ public function authenticate()
...
-        if (! $this->auth->byId($id)) {
-            return false;
+        $retrieveUser = $this->manager->getRetrieveUserOnAuthentication();
+
+        if ($retrieveUser && $this->auth->byId($id)) {
+            return $this->user();
         }
 
-        return $this->user();
+        return false;
```
tdhsmith: So when `'retrieve_user_on_authentication' => false`, this method will always do nothing and return false? My concern is that silently breaking a method is more confusing than instructing users not to use it. I'd prefer you to go all out and throw an exception instead (since the method won't ever be useful, you should alert those using it).

If that wasn't your intention or I'm misreading this, let me know. :wink:

tymondesigns:
After looking at this again, it has dawned on me that you can already get the behaviour you want, which as I understand it, is to check the validity of the token without worrying about pulling the authenticated user from the database.

So having said that, I think you could just use the pre-existing methods `check()` and `checkOrFail()`. Which will decode the token (thus validating it's `exp` and so on).

@jfoliveira Sound right?

jfoliveira:
@tymondesigns By using `check()` and `checkOrFail()` I would need to give up using middleware or Lumen auth guards as those already querying the database when a protected route is accessed.

So to use `check()` and `checkOrFail()` it seems the best way would be creating my own auth middleware, correct?

tymondesigns:
Yes I would definitely recommend adding your own middleware.

Let me know how it goes :+1:

unnaki:
@jfoliveira How did it go for you ? I'm looking for the same thing.

PR #471, opened 2016-01-30
Fixing jwt:secret command

w0rd-driven:
I apparently felt the need to live on the bleeding edge and immediately realized the jwt:secret command was broken. I would get the following exception:
`[Symfony\Component\Console\Exception\RuntimeException]  
Not enough arguments (missing: "show").`

It looks like the signature requires -- or - for commands. Changed the command to be --show and added an alias to -s. Also removed the word 'simply' to fall in line with other artisan commands.


tymondesigns:
Thanks!

Issue #239, opened 2015-09-09
Reponse Codes are not correct set


thenovacreator:
Hello Folks,

i got a bunch of issues right here. First of all i want to response with an 401 HTTP Code if no token is in the header, but i do not found a way to do this. After that i thought to just put an invalid Token if i do not have one to get this behavior. But on invalid tokens i get a HTTP 400 too.

Following your instructions to add this code:

        if ($e instanceof \Tymon\JWTAuth\Exceptions\TokenExpiredException) {
            return response()->json(['token_expired'], 401);
        } else if ($e instanceof \Tymon\JWTAuth\Exceptions\TokenInvalidException) {
            return response()->json(['token_invalid'], 401);
        }

to the Laravel error Handler gives me not what i want. As i figured out this render method is not even called (i jut putted a dd() in front of these lines). So i put my eyes on the JWT Middelware which i am using. In there the "JWTException" is used for invalid tokens ... shouldn't it be the "TokenInvalidException"? Anyway if found out that if this exception is thrown there is the HTTP 400 Code set. So if i am replacing:

$e->getStatusCode() with 401 i got what i want but i think it can not be the right way. Anyway i do think its the right behavior at all. Could someone fix this or tell me how to fix this on my own?

Thanks a lot ... Novacreator

tdhsmith:
While tokens _are_ authorization keys in a sense, their presence (`token_not_present`) and well-formedness (`token_invalid`) don't tell us anything about whether a user is actually authorized. Instead they are failures of communication, and fall under `HTTP 400 Bad Request`.

If you are interested in using non-standard status codes, your approach is fine. The provided middlewares are only considered examples, and you are encouraged to write your own versions as necessary. If you're looking for a lower-level solution, you could also override [`TokenInvalidException`](https://github.com/tymondesigns/jwt-auth/blob/master/src/Exceptions/TokenInvalidException.php) to have `$statusCode = 401`.

tymondesigns:
As @tdhsmith has said, the included middleware simply offers an example use case, and thus is definitely not a perfect fit for every application. So you should implement your own

PR #473, opened 2016-01-30
Shouldn't token_not_provided respond a 401 ?

clorichel:
given HTTP spec, request here is not a `400 Bad Request` but a `401 Unauthorized`

see https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

What do you think ?


tymondesigns:
This has been discussed a few times before, see #239 and https://github.com/tymondesigns/jwt-auth/issues/117#issuecomment-140553831

As i've said a few times before, implementing your own middleware is the way to go :). Thanks though

clorichel:
Oh... Quite sorry about that ! I'm afraid I haven't checked before...

A beer from me when I'll get to Manchester ;)

tymondesigns:
No problem at all :)

PR #485, opened 2016-02-03
Make setter fluid

lucasmichot:
In the entire project, only this setter is not fluid.
I suppsose this was not on purpose


tymondesigns:
yep, good catch

PR #493, opened 2016-02-03
Fix a few docblocks FQN and constant param spacing

lucasmichot:
`@param2\stype\2svaribale`

- fix missing arguments name
- spacing

### DIFF
--- a/src/Http/RouteParams.php
+++ b/src/Http/RouteParams.php
@@ -26,8 +26,7 @@ class RouteParams implements ParserContract
     /**
      * Try to get the token from the route parameters.
      *
-     * @param  \Illuminate\Http\Request
-     *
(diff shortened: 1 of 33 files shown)


lucasmichot:
Kind of Laravel

src/Http/RouteParams.php
```diff
@@ -26,8 +26,7 @@ class RouteParams implements ParserContract
     /**
      * Try to get the token from the route parameters.
      *
-     * @param  \Illuminate\Http\Request
-     *
+     * @param  \Illuminate\Http\Request  $request
```
tymondesigns: can we keep the line space here plz

tymondesigns:
in fact, it's fine i'll sort it, Thanks!

PR #446, opened 2016-01-27
Cleaner Carbon expressions

tdhsmith:
Just some updates that @lucasmichot pushed to develop that I wanted back on master while I was tinkering.

PR #494, opened 2016-02-03
Use the same styleci definition on master as the one used on develop

lucasmichot:
And avoid merging conflicts

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,14 +1,6 @@
 preset: laravel
 
 enabled:
-  - concat_with_spaces
   - unalign_double_arrow
 
-disabled:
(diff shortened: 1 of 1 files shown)


lucasmichot:
Indeed : https://styleci.io/analyses/864nko

GrahamCampbell:
All this has done is cause conficts.

tymondesigns:
I doubt I will be doing a normal merge from dev -> master tbh... 

The changes are vast at this point, and no need to go through that pain

lucasmichot:
@GrahamCampbell teh branches are already not synced, as some PR, see https://github.com/tymondesigns/jwt-auth/pull/446 for instance

PR #505, opened 2016-02-12
Fix all time issues in test

lucasmichot:
Test should never rely on `time()` but on `\Carbon\Carbon::setTestNow()` instead


tymondesigns:
Awesome! Thanks :smile:

PR #512, opened 2016-02-15
Use ::class notation

lucasmichot:
Use `::class` notation.

---

Also add missing `"illuminate/routing": "5.1.* || 5.2.*"` in `require-dev`


tymondesigns:
Thanks man!

Issue #510, opened 2016-02-15
Proposal : allows payload construction with \DatetimeInterface


lucasmichot:
I think it would make sense to allows the construction of the payload using either `int` for timestamps, either `\DatetimeInterface` (for the specific claims)

tymondesigns:
what exactly would you propose? I can't really think of a nice way to do it atm

lucasmichot:
Create an abstract TimestampClaim which will allow either timestamps either DatetimeInterface
Or simply modify the validator of the date related claims
In JWT the timestamp is immediately converted to a Carbon instance; so it wouldn't change much IMO

tymondesigns:
Ok i see what you mean.. ill throw something together

lucasmichot:
I'll make a WIP PR

PR #521, opened 2016-02-17
Let time related claims handle DateTimeInterface values with an abstract DatetimeClaim

lucasmichot:
Let time related claims handle DateTimeInterface values.

**Note:**
`return parent::validate($value);` is called because we keep the function description, and all all of them are slightly different.
If keeping them different is not so important, all `validate` functions can be removed from any class extending `\Tymon\JWTAuth\Claims\DatetimeClaim`

---

Solution for #510

### DIFF
--- a/src/Claims/DatetimeClaim.php
+++ b/src/Claims/DatetimeClaim.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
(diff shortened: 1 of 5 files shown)


tymondesigns:
closing in favour of #522

PR #522, opened 2016-02-17
Let time related claims handle DateTimeInterface values with a DatetimeTrait

lucasmichot:
Alternative for #521 with a trait.

**Note:**
This time we get rid of the original `validate` function description.

---

Solution for #510


tymondesigns:
Awesome, yea I much prefer this with the trait than #521 

:+1:

PR #527, opened 2016-02-17
Remove useless else

lucasmichot:
Remove useless else

### DIFF
--- a/src/Providers/Storage/Illuminate.php
+++ b/src/Providers/Storage/Illuminate.php
@@ -114,9 +114,9 @@ class Illuminate implements Storage
 
         if ($this->supportsTags) {
             return $this->cache->tags($this->tag);
-        } else {
-            return $this->cache;
         }
+
+        return $this->cache;
     }
 
     /**


src/Commands/JWTGenerateSecretCommand.php
```diff
@@ -59,9 +59,9 @@ public function fire()
                     file_put_contents($path, str_replace(
                         'JWT_SECRET='.$this->laravel['config']['jwt.secret'], 'JWT_SECRET='.$key, file_get_contents($path)
                     ));
-                } else {
-                    return $this->comment('Phew... No changes were made to your secret key.');
                 }
+
+                return $this->comment('Phew... No changes were made to your secret key.');
```
tymondesigns: This one is necessary. If you answer "yes" then it will still output as if no changes were made

tymondesigns: Guess I should take another look at that command test...

lucasmichot: My bad @tymondesigns
See: 600866c73448ad12b547f210d75fa7f79a68c655

tymondesigns:
:+1:

PR #532, opened 2016-02-18
Fixes ResponseFactory not being instantiable building BaseMiddleware

GiampaoloFalqui:
Fixes [Target [Illuminate\Contracts\Routing\ResponseFactory] is not instantiable while building [...]](https://github.com/laravel/framework/issues/12368#issuecomment-185792630)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor
 composer.lock
 build
+*.sublime*
--- a/src/Middleware/BaseMiddleware.php
+++ b/src/Middleware/BaseMiddleware.php
@@ -13,7 +13,6 @@ namespace Tymon\JWTAuth\Middleware;
 
 use Tymon\JWTAuth\JWTAuth;
 use Illuminate\Contracts\Events\Dispatcher;
-use Illuminate\Contracts\Routing\ResponseFactory;
 
 abstract class BaseMiddleware
 {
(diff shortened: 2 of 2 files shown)


dlucian:
I'm having the same issue after `composer update`. When we can have this merged into `master`?

GiampaoloFalqui:
It seems like today reverting this workaround fix, the library works again. Maybe something Illuminate's team forgot to add? I don't know. I'll close this since it's not needed anymore.

GiampaoloFalqui:
This is a workaround without touching the library to fix this issue (Lumen only, I think).

Add in your AppServiceProvider these three lines of code:
```php
$this->app->singleton('Illuminate\Contracts\Routing\ResponseFactory', function ($app)
{
    return new ResponseFactory($app['Illuminate\Contracts\View\Factory'], $app['Illuminate\Routing\Redirector']);
});
```

dlucian:
@GiampaoloFalqui Thanks for the fix, it's working great! 
I was already started worrying on how to deploy this with the library changes.

tymondesigns:
So this is a bug within Lumen then? If there is a workaround then I'll close this.

dlucian:
@tymondesigns Yeah, but might be worth mentioning this in the Installation instructions. Otherwise some people are going to go through some rough times :)

hopewise:
I have the same issue here, I am on lemon, but the above solution didn't work..

ihsanberahim:
@GiampaoloFalqui TQ! It works!

PR #542, opened 2016-02-26
Directly call \Illuminate\Support\Str::startsWith

lucasmichot:
Directly call `\Illuminate\Support\Str::startsWith`


tymondesigns:
:+1:

PR #551, opened 2016-03-01
Directly call \DateTime::getTimestamp

lucasmichot:
Directly call `\DateTime::getTimestamp`, no need to call `\Carbon\Carbon::__get`


tymondesigns:
Thanks! I've also just changed the other instances to match

PR #552, opened 2016-03-01
Abstract Provider

lucasmichot:
As only the boot methods differ, the service providers can be abstracted.

---

https://styleci.io/analyses/qord5g can be auto-applied after merge.


lucasmichot:
Ping @GrahamCampbell : https://styleci.io/analyses/8ADgBG
It seems strange that a final dot is added in this case.
`* {@inheritdoc }` => `* {@inheritdoc}.`

tymondesigns:
I'll merge this in, just need to add the exception to the phpunit config

tymondesigns:
I'll also remove those additions of the periods

lucasmichot:
:+1:

tymondesigns:
Looks like Graham quietly fixed this :smile: https://styleci.io/analyses/XNAJjE

PR #562, opened 2016-03-07
Update php version constraint

vinkla:
Instead of allowing all future versions of PHP we can be more specific.


tymondesigns:
:+1:

PR #568, opened 2016-03-09
Update \Tymon\JWTAuth\Validators\TokenValidator::validateStructure

lucasmichot:
Update `\Tymon\JWTAuth\Validators\TokenValidator::validateStructure` so that validation is stricter.
Also add more tests.

### DIFF
--- a/src/Validators/TokenValidator.php
+++ b/src/Validators/TokenValidator.php
@@ -36,8 +36,16 @@ class TokenValidator extends Validator
      */
     protected function validateStructure($token)
     {
-        if (count(explode('.', $token)) !== 3) {
+        $parts = explode('.', $token);
+
+        if (count($parts) !== 3) {
             throw new TokenInvalidException('Wrong number of segments');
+        }
+
+        $parts = array_filter(array_map('trim', $parts));
+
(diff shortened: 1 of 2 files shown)


src/Validators/TokenValidator.php
```diff
@@ -36,10 +36,18 @@ public function check($value)
...
-        if (count(explode('.', $token)) !== 3) {
+        $parts = explode('.', $token);
+
+        if (count($parts) !== 3) {
             throw new TokenInvalidException('Wrong number of segments');
         }
 
+        $parts = array_filter(array_map('trim', $parts));
+
+        if (count($parts) !== 3 || implode('.', $parts) !== $token) {
```
tymondesigns: Isn't the first part of this condition redundant? As it's already checked above

lucasmichot: Nope because, in this case we test the filtered array.

First test: `one..three` => 3 parts
Second test: `one.three` => 2 parts

tymondesigns:
Oh yea I see now.. (I was on a train is my excuse :smiley:) Thanks

PR #572, opened 2016-03-09
Use === null or !== null notation

lucasmichot:
Use === null or !== null notation


tymondesigns:
Why is this better?

tymondesigns:
I mean doesn't it yield exactly the same results

lucasmichot:
Increase performances (and also readability IMO)

tymondesigns:
OK I see it is marginally quicker. I have no readability preference here anyways

PR #582, opened 2016-03-10
Move \Commands to \Console

lucasmichot:
Respect Laravel app architecture and move \Commands to \Console


tymondesigns:
yep makes sense

PR #587, opened 2016-03-11
Move \Middleware to \Http\Middleware

lucasmichot:
Respect Laravel app architecture and move \Middleware to \Http\Middleware


tymondesigns:
Could you add this change to the service provider aswell?

https://github.com/tymondesigns/jwt-auth/blob/develop/src/Providers/LaravelServiceProvider.php#L14-L16

lucasmichot:
Done @tymondesigns

PR #588, opened 2016-03-11
Cache packages and use --prefer-dist

lucasmichot:
And speedup Travis builds

---

See https://github.com/composer/composer/issues/4884#issuecomment-195229989


tymondesigns:
Awesome thanks

PR #593, opened 2016-03-13
Fix some imports in AuthServiceProvider

brandonson:
Not sure how these ended up missing, but they weren't in the most recent commit and that's causing load failures.


tymondesigns:
Thanks!

PR #613, opened 2016-03-18
Fix , => typo in LumenServiceProvider

BrettTrotter:
Otherwise this seems to yield

screen shot 2016-03-18 at 10 01 33 am


src/Providers/LumenServiceProvider.php
```diff
@@ -32,7 +32,7 @@ public function boot()
         $this->mergeConfigFrom($path, 'jwt');
 
         $this->app->routeMiddleware([
-            'jwt.auth', Authenticate::class,
+            'jwt.auth' => Authenticate::class,
```
BrettTrotter: Otherwise this seems to yield

screen shot 2016-03-18 at 10 01 33 am

tymondesigns:
good catch! :+1:

PR #619, opened 2016-03-22
Add Ability To Set Infinite Refresh TTL

MitchellMcKenna:
Similar to how never expiring tokens is currently supported, adds support for infinite refresh TTL time. 

Just like how you can set `JWT_TTL=null`, adds support to set `JWT_REFRESH_TTL=null`.

Opposed to using never expiring tokens, a mobile app developer may want the additional security of having tokens expire after some time (say, in case it is compromised in transit somehow, at least it may expire before it could be used maliciously), but allows the mobile app to refresh the token with the server at any time after it expires.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -112,6 +112,11 @@ return [
     | the original token being created until they must re-authenticate.
     | Defaults to 2 weeks.
     |
+    | You can also set this to null, to yield an infinite refresh time.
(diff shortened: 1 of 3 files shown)


tymondesigns:
Makes sense, thanks!

iamrutvik:
how would you set never expiring tokens ?

PR #620, opened 2016-03-23
Test all common \Tymon\JWTAuth\Claims\Factory fields

lucasmichot:
Test all common `\Tymon\JWTAuth\Claims\Factory` fields


tymondesigns:
:+1:

PR #621, opened 2016-03-23
Add tests for setHeaderPrefix and setHeaderName

lucasmichot:
Add tests for setHeaderPrefix and setHeaderName


tymondesigns:
Thanks!

PR #622, opened 2016-03-23
Only calculate default values when this is required

lucasmichot:
Only calculate default values when this is required


tymondesigns:
curious.. what's the rationale for this?

tymondesigns:
just given that this is the code that runs:

```php
public function get($key, $default = null)
{
    return array_key_exists($key, $this->parameters) ? $this->parameters[$key] : $default;
}
```

I guess a very slight perf increase?

lucasmichot:
The fact is that we calculate the default in all cases.
Using ?: avoids that

tymondesigns:
I see your point :+1:

PR #630, opened 2016-03-24
Make QueryString parser do what it says it does, and add a new InputS…

ashleigh-hopkins:
…ource parser to fulfil the role of the current QueryString parser

PR #635, opened 2016-03-29
Remove useless header set

lucasmichot:
Remove useless header set


GrahamCampbell:
Indeed. This is not be be confused with the accept header. ;)

lucasmichot:
Yep

tymondesigns:
Whoops, should've caught that myself in #630. Thanks for that

PR #636, opened 2016-03-29
Also pass parent Exception and parent Exception code

lucasmichot:
Also pass parent Exception and parent Exception code


tymondesigns:
Cheers :+1:

PR #639, opened 2016-03-29
fix typo

pjona:


LukasMa:
@tymondesigns I think this can be closed, since that typo was fixed with #638 right?

Issue #654, opened 2016-04-05
Should timestamps be returns as numbers instead of strings?


mottihoresh:
I came across this issue while trying to implement jwt authentication on iOS, it seems like the open source decoders are expecting those values to be number

tymondesigns:
we'll continue the discussion at #655

PR #655, opened 2016-04-05
added JSON_NUMERIC_CHECK to json_encode

mottihoresh:
Based on the JWT specs, the expiration date should be returned as a number. I've created an issue #654 describing an issue I had trying to use the token generated within a iOS app where it would reject the exp since it wasn't a number.

The proposed fix should fix the issue.

### DIFF
--- a/src/Payload.php
+++ b/src/Payload.php
@@ -146,7 +146,7 @@ class Payload implements ArrayAccess, Arrayable, JsonSerializable, Jsonable, Cou
      */
     public function __toString()
     {
-        return $this->toJson(JSON_UNESCAPED_SLASHES);
+        return $this->toJson(JSON_NUMERIC_CHECK & JSON_UNESCAPED_SLASHES);
     }
 
     /**


tymondesigns:
How are you creating the token? because this is simply the method that returns the value when casting the object to a string.

Also this shouldn't even be an issue since it was addressed quite a while ago - The Factory class returns the values as integers since it uses [`DateTime::getTimestamp()`](http://php.net/manual/en/datetime.gettimestamp.php)

I'm not opposed to adding this in, but it should be:

```php
return $this->toJson(JSON_NUMERIC_CHECK | JSON_UNESCAPED_SLASHES);
```

mottihoresh:
@tymondesigns  it seems like I was an older version, had in my composer file 0.5.0 instead of 0.5.*, when i retested it it seems to be coming back ok now. (v 0.5.9)

tymondesigns:
@mottihoresh no worries 👍

PR #657, opened 2016-04-08
Tests \Tymon\JWTAuth\Support\CustomClaims::customClaims

lucasmichot:
Test `\Tymon\JWTAuth\Support\CustomClaims::customClaims`, not its alias.

---

Tests failing for another reason


tymondesigns:
fixed tests now 👍 +

PR #662, opened 2016-04-12
Add return void docblock

lucasmichot:
Return void was sometimes set for `__construct`, sometimes not.
This PR sets it everywhere to respect Laravel coding style

---

**Note:** in some cases there is an empty line before `@return`, in some other not.
StyleCI does not find anything wrong - but IMO, it think it should

### DIFF
--- a/src/Blacklist.php
+++ b/src/Blacklist.php
@@ -44,6 +44,8 @@ class Blacklist
 
     /**
      * @param  \Tymon\JWTAuth\Contracts\Providers\Storage  $storage
+     *
+     * @return void
      */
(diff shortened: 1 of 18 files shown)


tymondesigns:
need that extra line break before the returns

lucasmichot:
On my way to the airport now - cannot commit
Might be worth adding phpdoc_separation fixer then

On 13 April 2016 at 13:38:28, Sean Tymon (notifications@github.com) wrote:

need that extra line break before the returns

tymondesigns:
@lucasmichot good call.. didn't know about that one :)

PR #663, opened 2016-04-13
Update branch-alias to 1.0-dev

barryvdh:
You're building the 1.0 version on the develop branch now, right? Or are you still planning to release 0.6?
In that case, change the branch-alias to 1.0, so people can require `1.x@dev` and get the latest dev version (instead of the alpha)


tymondesigns:
Thanks for that :+1:

PR #676, opened 2016-04-16
Update phpunit config

vinkla:
We shouldn't allow risky tests.


tymondesigns:
@vinkla I've actually added this now, with the exception of failing on tests that don't make any assertion, so cheers for that :)

vinkla:
What you could do is to `assertTrue(true);` on those specific tests if you want to add it.

PR #679, opened 2016-04-18
Update composer.json

jycr753:
This version of jose passes all unit testing

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "illuminate/contracts": "5.1.* || 5.2.*",
         "illuminate/http": "5.1.* || 5.2.*",
         "illuminate/support": "5.1.* || 5.2.*",
-        "namshi/jose": "^6.0",
+        "namshi/jose": "^7.0",
         "nesbot/carbon": "^1.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -24,7 +24,7 @@
         "illuminate/contracts": "5.1.* || 5.2.*",
         "illuminate/http": "5.1.* || 5.2.*",
         "illuminate/support": "5.1.* || 5.2.*",
-        "namshi/jose": "^6.0",
+        "namshi/jose": "^7.1",
```
lucasmichot: `"^7.0"` is enough

jycr753: enough? is there a good reason why not update to the latest?

lucasmichot: People might already use `7.0.0` in their project - forcing to use `^7.1` would cause package conflicts for them.

lucasmichot:
You could probably just reset to 2eb7c442e2b6a9e51f9aa6eabb67056e7b0094d1 and `push --force` it, as https://github.com/tymondesigns/jwt-auth/pull/679/commits/b402b71607f370ad8c59afa37011a5effe5427fb just reverses https://github.com/tymondesigns/jwt-auth/pull/679/commits/2eb7c442e2b6a9e51f9aa6eabb67056e7b0094d1

jycr753:
Yeah sorry just created another one with the right things, sorry about that

PR #682, opened 2016-04-18
Skip token validation when user is already set

TheGIBSON:
There is a situations where we need to set user manually and internally dispatch routes that uses jwt auth without token validation

### DIFF
--- a/src/Http/Middleware/Authenticate.php
+++ b/src/Http/Middleware/Authenticate.php
@@ -27,7 +27,9 @@ class Authenticate extends BaseMiddleware
      */
     public function handle($request, Closure $next)
     {
-        $this->authenticate($request);
+        if (! $this->auth->userExists()) {
+            $this->authenticate($request);
+        }
 
         return $next($request);
     }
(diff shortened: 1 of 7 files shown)


src/Http/Middleware/Authenticate.php
```diff
@@ -27,7 +27,9 @@ class Authenticate extends BaseMiddleware
      */
     public function handle($request, Closure $next)
     {
-        $this->authenticate($request);
+        if (! $this->auth->userExists()) {
```
jycr753: Does the user exist in this level? dont you need to get it from $guard?

TheGIBSON: If we register:
```PHP
    public function register()
    {
        $this->app->singleton(\Tymon\JWTAuth\Providers\Auth\Illuminate::class, function ($app) {
            return new \Tymon\JWTAuth\Providers\Auth\Illuminate(
                $app[\Illuminate\Contracts\Auth\Factory::class]->guard('user')
            );
        });
    }
```
Then the user will exist in this level. If not, it will be authenticated as before.

TheGIBSON:
ok, i will add ASAP

TheGIBSON:
done

lucasmichot:
Can you rebase @TheGIBSON ?

PR #697, opened 2016-05-01
Also tests with PHP > 5.5.9

lucasmichot:
Also tests with PHP > `5.5.9`

---

https://travis-ci.org/tymondesigns/jwt-auth/jobs/126997619 and https://travis-ci.org/tymondesigns/jwt-auth/jobs/126997620 pass with PHP `5.5.21`


tymondesigns:
👍

PR #699, opened 2016-05-02
Add `check()` method to the guard

SaeidRp:
without this method we always get `true` when calling `Auth::setToken($token)->check()`


tymondesigns:
Thanks but you can already call the check method (see the `__call` magic method)

Issue #719, opened 2016-05-20
Refresh Token Error Array to string conversion


Abdulhmid:
I use refresh token but any error like this "PayloadFactory.php line 194:
Array to string conversion "

tdhsmith:
How are you initially creating the token? For some reason your `sub` claim is being stored as an array, and I think it must be happening earlier than the actual call to `refresh`...

Abdulhmid:
I create token like this ( with array parameter ),
 $payload = JWTFactory::make($array);
 return JwtAuth::encode($payload)->get();

tdhsmith:
Hrm that looks fine. And what does your `$array` look like? The `sub` value isn't multidimensional is it?

tdhsmith:
Aside: @tymondesigns there's still some odd array manipulation going on in the required claims checks in v0.5.9 at `PayloadFactory->buildClaims` and `PayloadValidator->validateStructure` (cf. 70f4d2eec). The first is just doing nothing and probably wouldn't affect anybody (except in the extreme edge case of using low integers as your claim names). The second though isn't checking that required claims are present anymore... it effectively just checks that the payload has at least as many claims as there are required claims.

tdhsmith:
Actually thinking about it, the first edge case could be pretty bad if someone accidentally passes in an unkeyed array to `PayloadFactory::make`. Then basically none of their custom claims will get added. It's still a user error of incorrect arguments, but it would be very confusing for them to debug...

I'll submit a fix sometime tomorrow.

PR #738, opened 2016-06-03
Fix claims array comparisons

tdhsmith:
Issue in 0.5 that I mentioned in #719.

- Simplify `PayloadFactory` addition of custom claims and prevent issue with numerically-keyed claims.
- Fix `PayloadValidator` required claims check, which was only verifying that there were at least as many claims as requireds and not comparing keys.
- Create tests for both scenarios.


tymondesigns:
yea makes sense, cheers!

Issue #351, opened 2015-12-16
JWTAuth refresh


igostavro:
Getting an exception when trying to refresh a token manually or via the middleware:

```
ErrorException (E_NOTICE)

Array to string conversion
    <path>/tymon/jwt-auth/src/PayloadFactory.php
         * @return string
         */
```
```php
    protected function jti()
    {
        $sub = array_get($this->claims, 'sub', ''); <--- sub is an array after this line
        $nbf = array_get($this->claims, 'nbf', '');
 
        return md5(sprintf('jti.%s.%s', $sub, $nbf)); <--craps out here
    }
``` 
the debugger shows that $this->claims["sub"] is an array

Laravel Framework version 5.1.26 (LTS)
In my composer.json:
"tymon/jwt-auth": "0.5.*"

Thanks in advance!

tdhsmith:
That's pretty strange. Did you ever figure this out? Was `sub` an array before that point?

You might be able to coerce the payload factory into building a token with an array-valued claim (probably easier in 0.5 than 0.6 anyway), but it's strange that this would only happen on refresh. Maybe Namshi parses strings into arrays if they're JSON style? That would be odd too...

tdhsmith:
In any case, if you have an example token that does this, it might help.

tymondesigns:
@igostavro did you resolve this?

igostavro:
Never seemed to happen again. Not sure what caused it.

Thanks for the help though:)

Sent from my iPhone

On Mar 10, 2016, at 4:19 AM, Sean Tymon <notifications@github.com<mailto:notifications@github.com>> wrote:


@igostavro<https://github.com/igostavro> did you resolve this?

-

Abdulhmid:
i'm still error like this "Array to string conversion", anybody can help me ,

Abdulhmid:
@igostavro  can u help me ?

Abdulhmid:
@tymondesigns ?

anasinfo:
aslam 
what help do u want ?? in laravel??ahmed
[...]

PR #739, opened 2016-06-03
Bailout to json_encode when building jti claims

tdhsmith:
The default `jti` would stumble on unstringable `sub` values, which were occasionally happening for some users.

Addresses #719 and #351 and another one I can't find. Or rather it fixes a _symptom_ of those issues. I don't think anyone has identified why they had a non-standard subject value (and whether it was their app or some edge case in the library), but there isn't a good reason `jti` should choke on that scenario.


tymondesigns:
Yea this is a good thing to catch, thanks!

PR #743, opened 2016-06-05
Let the claim implements Arrayable

lucasmichot:
As we already have `\Tymon\JWTAuth\Claims\Claim::toArray`


tymondesigns:
cheers dude 👍

PR #744, opened 2016-06-05
Add matches function to Payload

lucasmichot:
Add matching functions

### DIFF
--- a/src/Payload.php
+++ b/src/Payload.php
@@ -73,10 +73,10 @@ class Payload implements ArrayAccess, Arrayable, JsonSerializable, Jsonable, Cou
             return false;
         }
 
-        $claims = new Collection($this->getClaims()->toArray());
+        $claims = $this->getClaims();
 
         foreach ($values as $key => $value) {
-            if ($claims->where($key, $value, $strict)->isEmpty()) {
+            if (! $claims->has($key) || ! $claims->get($key)->matches($value, $strict)) {
                 return false;
             }
         }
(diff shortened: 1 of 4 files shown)


src/Payload.php
```diff
@@ -60,6 +60,41 @@ public function getClaims()
...
+     * @return bool
+     */
+    public function matches(array $values, $strict = false)
+    {
+        if (empty($values)) {
+            return false;
+        }
+
+        foreach ($values as $key => $value) {
+            if (($strict && $this->get($key) !== $value) || $this->get($key) != $value) {
```
tymondesigns: doesn't this negate the point of having a strict mode?

maybe something like this would be better? - https://github.com/illuminate/support/blob/5.2/Collection.php#L222-L228

lucasmichot: @tymondesigns I've changed things a bit - nevertheless when I see #746 (on which I'll have a deeper look), I feel like the claim itself should be responsible for this strict/not-strict match on itself

Let me know you thoughts on that !

lucasmichot: @tymondesigns Let me know about this new approach added by https://github.com/tymondesigns/jwt-auth/pull/744/commits/6b6d52b856061e31fd6a6a9331f4c15710233fd5 ?

tymondesigns:
Yea looks good to me 👍

lucasmichot:
Not WIP anymore

tymondesigns:
Cool, thanks!

Issue #735, opened 2016-06-01
Proposal: Handle validation, defaults, & refreshes in individual Claim classes


tdhsmith:
I've been thinking it might be cleaner and more modular to hold all of the validation methods and defaults inside each of the `Claim` classes. This would make claim initialization and validation a lot more configurable, and checks would be more internalized so the user doesn't have to manually examine the token payload.

Before I go too far, I want to highlight a couple use cases:
- I'd like to have `iss` verify that it matches the current domain as a sanity check. (Really this shouldn't come up in practice because different issuers would normally have different keys, but it's possible some people want multiple token issuing services all running on the same Laravel instance.)
- I'd like to implement an `csrf` claim to help mitigate request forgery when storing JWTs in cookies. A couple of existing services already use a claim for this (Stormpath [has an `xsrfToken` claim](https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage#crayon-574e1259c2eaa543356369) and OpenID Connect [has a `nonce` claim](http://openid.net/specs/openid-connect-core-1_0.html#IDToken)). I could certainly do this already, but it would be nice if I didn't have to expose the JWT internals to do it.
- People have previously requested different token refresh flows, like being able to chain refreshes infinitely but still having an expired-but-refreshable period (not currently possible since the period is always determined from the original `iat`) or having a finite number of refreshes (we could have a `refresh_count` claim that tracked this). As it stands these would require a lot of conditional code inside the manager/validator/factory and extra config values, but with claim handling it would just take a new claim class.

Here's an approximate structure as an example:

``` php

abstract class Claim
{
    protected $value;
    public $name; // tells the manager which claim keys to map to this class

    // Faked "enumerated" value for action in case there end up being others
    const CHECK = 0;
    const REFRESH = 1;

    public function __construct($value = null) {
        if (is_null($value)) {
            // if we have no args in our constructor, use the default
            $this->reset();
        } else {
            $this->set($value);
        }
    } 

    public function validate($action = self::CHECK) {
        switch ($action) {
            case self::CHECK:
                return $this->validateCheck();
            case self::REFRESH:
                return $this->validateRefresh();
        }
        throw new \Exception("Invalid action flag.");
    }

    abstract function validateCheck();
    abstract function validateRefresh();

    final public function getInternal() {
        return $this->value;
    }
    final public function setInternal($newValue) {
        $this->value = $newValue;
    }

[...]

Issue #706, opened 2016-05-06
1.0.0 Alpha - support of binary uuids


autaut03:
Hey! I'm using your package, but right now I have to replace your auth providers just to modify $id used in auth provider byId method.. Really, it would be just great if you add some function to your JWTSubject interface like transformIdentifierIn and transformIdentifierOut.. and why do we need this - because I personally use UUID's and store it as binary 16. So when I try to authenticate - it gives me error cuz it's unreal to find a user with string'ed uuid.

Unfortunatelly there's no normal way of auto-transforming UUID's in where clause so I have to do it manually (almost) everywhere.

helloflow:
+1 I'm facing the same issue. Would be great if their is a solution for working with UUID's.

dodohack:
@autaut03 Have you resolved this issue? I have exactly the same problem as yours. Just trying to use binary(16) uuid instead of auto-incrementing id in user table.

autaut03:
@dodohack Sure. Replaced AuthProvider's method and that's all. Not an elegant solution but it works.

PR #746, opened 2016-06-05
Refactor [feedback wanted]

tymondesigns:
@tdhsmith @lucasmichot and others!

I would love to get your initial thoughts on the approach I have taken here...

This is by no means ready for a merge! but I think it's a step in the right direction. There is much more work to do on this - (the tests are nowhere near green and there's bound to be some bugs).

As I say, just wanted to get some thoughts on the drastic change. I was inspired by #735 to begin this, so just want some feedback in case I have introduced any obvious flaws with this approach.

Essentially, Claims are now responsible for the validation of themselves within various contexts.. (creation, when checking the validity of a payload, and refreshing).

And the Claim factory is now responsible for creating the initial Claim instances (except the `jti`).

And lots more tweaks...

I welcome your comments!

Any further clarifications, let me know

Thanks


tdhsmith:
Looking good so far! I think I'm going to comb through this a bit later and play extreme-modularization advocate though. 😛  

On that note already: Are you considering moving the default-setters ('iss()', 'nbf()' etc in `Factory`) inside each claim class? (e.g. as `Claim::setDefault`)

lucasmichot:
I do agree with @tdsmith - good so far

src/Claims/Claim.php
```diff
@@ -98,15 +93,37 @@ public function getName()
...
+    }
+
+    /**
+     * Validate the Claim within a refresh context.
+     *
+     * @param  int  $refreshTTL
+     *
+     * @return bool
+     */
+    public function validateRefresh($refreshTTL)
```
tdhsmith: Is every possible refresh flow going to use `$refreshTTL` and not need something else? If there are other possibilities, I'd rather leave out the argument and do something more generic to get those kinds of values. Maybe inject the `Factory` instance in the `Claim` constructor, and have a general `Factory::getSetting` method providing access to configured values like `ttl` & `refreshTTL`.

Of course, I'm not sure if there _are_ legitimate cases for needing other parameters in the refresh flow. Even if you have the chaining function differently, it would still be based on a time delta. In #735 I mentioned a possible `RefreshCount` claim that would need a `refresh_limit` value, but that's a fringe idea and I'm not actually certain why anyone would want that limitation over a timed one. (Though I believe it was requested in the issues here once before)

src/Factory.php
```diff
@@ -127,31 +119,37 @@ protected function buildClaims()
 
         // add the default claims
         foreach ($this->defaultClaims as $claim) {
-            $this->addClaim($claim, $this->$claim());
+            $this->addClaim($claim, $this->claimFactory->make($claim));
         }
 
         // add custom claims on top, allowing them to overwrite defaults
         $this->addClaims($this->getCustomClaims());
```
tdhsmith: Feature: have custom claims use the claim factory as well.

The custom claims in this factory don't necessarily mean "custom" in the sense of the `Custom` claim. The user might still have defined a special claim class with validation and behavior. So in that sense they're more like "additional" claims beyond the default set, and it would be nice if there were a way to still run through their classes' constructors/methods.

Additionally, if you support moving the defaults into the claims, you could allow unkeyed array entries that are just claim names. So this would work:
```php
Factory::customClaims(['name' => 'myApplication', 'cnf'])->make();
// if those exist in the class map, would result in claim Factory running:
$classMap['name']->setValue('myApplication');
$classMap['cnf']->setDefault();
```
(Although technically detecting that probably requires removing support for numerically-named claims.)

src/Factory.php
```diff
@@ -127,31 +119,37 @@ protected function buildClaims()
...
 
+        // add the jti last since it is based on all other claims,
+        // but only if it hasn't already been added.
+        if (! $this->claims->has('jti')) {
+            $this->addClaim('jti', $this->jti());
```
tdhsmith: Possibility: `jti` doesn't need to be a special case.

If the claim factory handles construction of the entire collection (independent question: is there a reason to leave that in the main Factory?), then we could do:
```php
abstract class Claim {
    public static $creationPriority = 0;
    // ...
}
class JwtId extends Claim {
    public static $creationPriority = -100;
    public function setDefault() {
        $this->value = md5(sprintf('%s.%s', $this->factoryInstance->claims->toJson(), Str::quickRandom()));
    }
    // ...
}
class ClaimFactory {
    public function getCollection(array $rawClaims)
    {
        return $collect($rawClaims)->sortByDesc(function ($value, $name) {
            // sort the collection from highest to lowest priority
            return self::$classMap[$name]::$creationPriority;
        })->transform(function ($value, $name) {
            // replace each entry with the claim instance
            return $this->get($name, $value);
        });
    }
    // ...
}
```

That would allow us to guarantee order of creation, so `jti` could still depend on the other values, but avoid a hard-coded special case.

Regardless, I think it might be good to consider letting the claim factory build the whole collection. Just have regular factory handle the way the user provides input (collecting the claims as an array until `make` is called).

tymondesigns:
@tdhsmith 

> On that note already: Are you considering moving the default-setters (iss(), nbf() etc in Factory) inside each claim class? (e.g. as Claim::setDefault)

I like the idea, but I can't think of a clean way to do it structurally, simply because each Claim has different requirements, e.g. `exp` needs ttl and `iat` needs the refresh ttl etc. Do you have any ideas on that? I guess like you mentioned, there could be a global Settings Class to provide these values statically... maybe.

> The custom claims in this factory don't necessarily mean "custom" in the sense of the Custom claim. The user might still have defined a special claim class with validation and behavior. So in that sense they're more like "additional" claims beyond the default set, and it would be nice if there were a way to still run through their classes' constructors/methods.

So it should be already possible to provide a custom claim instance to the PayloadFactory, and have it run through the validations and everything, because of the following method where it checks the type:

```php
protected function resolveClaims()
{
    return $this->claims->map(function ($value, $name) {
        return $value instanceof Claim ? $value : $this->claimFactory->get($name, $value);
    });
}
```

is that what you meant?

> Possibility: jti doesn't need to be a special case.

> If the claim factory handles construction of the entire collection (independent question: is there a reason to leave that in the main Factory?), then we could do:

You could be right there. I like the idea of the priority, So i'll try that one out shortly!

> Regardless, I think it might be good to consider letting the claim factory build the whole collection. Just have regular factory handle the way the user provides input (collecting the claims as an array until make is called).

I do like the sound of that... I'll try this out and see how it looks

> Is every possible refresh flow going to use $refreshTTL and not need something else? If there are other possibilities, I'd rather leave out the argument and do something more generic to get those kinds of values. Maybe inject the Factory instance in the Claim constructor, and have a general Factory::getSetting method providing access to configured values like ttl & refreshTTL.

Yea I'm not totally happy with how this is currently set up, I'll see what I can cook up.

Thanks for the feedback! 👍

lucasmichot:
@tymondesigns I am feeling more and more that this package should be splitted in two different packages:

1. The JWT generation, encoding and decoding
2. The authentication part

src/Claims/Collection.php
```diff
@@ -0,0 +1,109 @@
...
+use Illuminate\Support\Collection as IlluminateCollection;
+
+class Collection extends IlluminateCollection
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($items = [])
+    {
+        parent::__construct($this->sanitizeClaims($items));
```
lucasmichot: Maybe move this call to `\Illuminate\Support\Collection::getArrayableItems` ?

src/Claims/Collection.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Determine of the Collection contains all of the given keys.
+     *
+     * @param  array  $claims
+     *
+     * @return bool
+     */
+    public function hasAllClaims($claims)
+    {
+        return count(array_diff(array_keys($this->toArray()), $claims)) === 0;
```
lucasmichot: I feel `\Illuminate\Support\Collection` functions can be used there

src/Claims/DatetimeTrait.php
```diff
@@ -35,50 +36,38 @@ public function setValue($value)
...
-     * @return bool
+     * {@inheritdoc}
      */
-    protected function checkNumeric($value)
+    public function validateCreate($value)
     {
-        return is_numeric($value);
-    }
+        if (! is_numeric($value)) {
+            throw new InvalidClaimException('Invalid value provided for claim "'.$this->getName().'": '.$value);
```
lucasmichot: Why not a new `InvalidClaimValueException` that direct take the claim as an argument ?

tymondesigns: Why didn't I think of that 😄

src/Claims/Collection.php
tymondesigns: Yea I tried and gave up in the end.. will look at it again though

src/Claims/Collection.php
tymondesigns: Good shout!

tymondesigns:
@lucasmichot I also think it would be better to separate out the package too.. I just need to rationalise what needs to be separated out and what doesn't and map out what the requirements would be of each package. I'd love to make the JWT portion totally framework agnostic if I can

src/Claims/DatetimeTrait.php
lucasmichot: 👍

src/Claims/Collection.php
lucasmichot: `return $this->keys()->diff($claims)->empty()`; should do the trick

tdhsmith:
> So it should be already possible to provide a custom claim instance to the PayloadFactory, and have it run through the validations and everything, because of the following method where it checks the type: [...]

I guess what I meant was that the claim factory would hold entries in its map for custom classes. Then in the payload factory, users don't have to deal with the classes directly -- they can just use the claim name.

So you'd define extra mappings (either in config, or somewhere on application boot). Afterward, you can just add that claim by adding [`'customClaimName' => 'argForConstructor']`.

tdhsmith:
Just to document another case where encapsulation would be useful: see issue #706. Instead of futzing with `AuthProvider` or their suggestion of `JWTSubject` changes, it could just be adapted with a subclass of `Subject`:
```php
class UUIDSubject extends Subject {
    public function setValue($value)
    {
        $this->validateCreate($value);
        $this->value = hex2bin($value);
        return $this;
    }
    public function getValue()
    {
        return hex2bin($this->value);
    }

}
```

(Though I might be misunderstanding exactly which format the UUIDs are on the backend.) Then if they set that as their default class for `sub` claims, everything would be peachy.

tdhsmith:
Ooh, two more thoughts on how this might streamline the rest of the library:

 1. The `fromUser`/`toUser` stuff ican be removed from core in favor of subject claim subclasses:
```php
class EloquentPrimaryKeySubject extends Subject {
    protected $modelClass; // set in the constructor or in a setter or something (not shown)
    public function validateCreate($user)
    {
        if (! is_a($user, Model::class)) {
            throw new InvalidClaimException($this);
        }
    }
    public function setValue($user)
    {
        $this->validateCreate($value);
        $this->value = $user->getKey();
        return $this;
    }
    public function getValue()
    {
        return call_user_func([$this->modelClass, 'findOrFail'], $this->value);
        // alternatively, do this stuff via Auth::loginUsingId(), if the "auth-encapsulating" approach is preferred
    }
}
```
 2. The blacklist caching times could be determined by the claims themselves as well. Something like an `BlacklistCacheTimer` interface on some timestamp claims requiring `getBlacklistTimeout()`, and the blacklist takes the maximum of all the values it receives and uses that for the cache.

PR #754, opened 2016-06-10
Replace CRLF with PHP_EOL

KarboniteKream:
Uses correct line endings on the target platform.

### DIFF
--- a/src/Console/JWTGenerateSecretCommand.php
+++ b/src/Console/JWTGenerateSecretCommand.php
@@ -49,7 +49,7 @@ class JWTGenerateSecretCommand extends Command
 
             // check if there is already a secret set first
             if (! Str::contains(file_get_contents($path), 'JWT_SECRET')) {
-                file_put_contents($path, "\r\nJWT_SECRET=$key", FILE_APPEND);
+                file_put_contents($path, PHP_EOL."JWT_SECRET=$key", FILE_APPEND);
             } else {
 
                 // let's be sure you want to do this


GrahamCampbell:
:-1:

GrahamCampbell:
Should be platform independent.

KarboniteKream:
What do you mean? It was using `\r\n` before.

tymondesigns:
Yea so `PHP_EOL` is platform agnostic, I think @GrahamCampbell may have misread the diff ?

GrahamCampbell:
No, I didn't missread this. PHP_EOL was not intended for that use.

GrahamCampbell:
> Yea PHP_EOL is platform agnostic,

No it's not. The whole point of it is that it's the exact opposite. It changes per OS. We have no need for that.

GrahamCampbell:
TBH, I'm actually just recommend changing it to use only the unix style, as is the PSR standard.

tymondesigns:
> TBH, I'm actually just recommend changing it to use only the unix style, as is the PSR standard.

OK are you referring to [PSR-2](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#22-files) here?

tymondesigns:
Also PHP_EOL will out put the correct output if on a *nix OS anyway right.. so does it matter?

GrahamCampbell:
But it'll output windows style on windows. Being on windows doesn't mean you have to use the windows style newlines. I don't.

PR #761, opened 2016-06-16
Added an exception to JWTGuard->user()

JaffParker:
Made `JWTGuard->user()` throw an exception instead of simply returning nothing. If it returns null, it's too annoying to check everywhere. An exception thrown would avoid that.


lucasmichot:
A extended JWT exception would make more sense IMO
Tests are failing.

JaffParker:
Yeah, makes sense. I didn't check the tests. I'll work on this edit a little more.

PR #762, opened 2016-06-17
Made the JWTGuard->user() method throw an exception in case of failure

JaffParker:
Before the method used to return `null` which was a cause of frustration every time someone wanted to just do `Auth::user()->email` (for example). Instead, we had to perform checks every time we call that function. With the exception thrown instead, we can ensure that nothing unexpected happens and keep the code clean.

Note: I wasn't sure how to name the exception the best way, so change it to whatever you think is better.

### DIFF
--- a/src/Exceptions/UserNotDefinedException.php
+++ b/src/Exceptions/UserNotDefinedException.php
@@ -1,0 +1,17 @@
+<?php
+
+/*
+ * This file is part of jwt-auth.
+ *
+ * (c) Sean Tymon <tymon148@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Tymon\JWTAuth\Exceptions;
+
+class UserNotDefinedException extends JWTException
+{
+    //
+}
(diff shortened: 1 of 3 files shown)


tymondesigns:
Although I understand your use case for this.. It will go against the [laravel guard interface](https://laravel.com/api/5.2/Illuminate/Contracts/Auth/Guard.html)

I think this should be a separate method like `Auth::userOrFail();` or something.

JaffParker:
That is a valid point, I'll look into it.

tymondesigns:
I think just one more quick test case for the happy path and we should be golden

JaffParker:
Well, I definitely gotta up my testing skill, the coverage still is down... I created a test case specifically for the method.

JaffParker:
Well, nevermind, apparently it's good... Let me know if you'd like something else adjusted.

lucasmichot:
Can you squash your commits?

JaffParker:
I hope that's what you meant. I squashed it all into one commit, but the told commits still stayed.

tymondesigns:
It's ok github now allows squashing when merging a PR

MitchellMcKenna:
Thank you so much for adding this, was having an issue where deleted users were causing errors because their auth token was still active in Redis but their user accounts were not in the database.

PR #790, opened 2016-07-12
Give secret generation command a `--force` option

tdhsmith:
As it stands you cannot generally call `JWTGenerateSecretCommand` from code because it will hit a user confirmation dialog when replacing an existing key, and as far as I can tell, there isn't an easy way to respond to artisan command dialogs from within the application. Providing a `--force` option allows us to skip this dialog for in-app command calls (it can also simplify build scripts that might be regenerating keys, so they no longer need to pipe in a "y").


tymondesigns:
Thanks!

PR #796, opened 2016-07-18
Directly access \Tymon\JWTAuth\Providers\JWT\Provider::$keys

lucasmichot:
No need to call `\Tymon\JWTAuth\Providers\JWT\Provider::getKeys`


tymondesigns:
Thanks! 👍

PR #823, opened 2016-08-12
Add set/get secret to provider

antonioreyna:
Adds the ability to set a custom secret for every JWT Token.


tymondesigns:
Thanks!

PR #826, opened 2016-08-14
Change $token = JWTAuth::getToken() to $token = JWTAuth::getToken()->get() on Wiki Page

laxx:
In the authentication section of the Wiki, it states that to get the token for the current request, call $token = JWTAuth::getToken(). But the getToken() method returns an instance of the Token class and using on a function like compact() would return {} since its not a variable but an object.


tymondesigns:
this contains no new code?

laxx:
No i was just making an observation on the info giving on the wiki page.

PR #828, opened 2016-08-15
ID's should be accessed through accessors

godlevel:
Original file tries to access User IDs directly. I propose they should be accessed through accessors.


tymondesigns:
Thanks but this is a non-issue on the next release

PR #840, opened 2016-09-05
Made compatible with phpunit tests when testing multiple tokens in on…

CodeHunger:
I did have some trouble testing with phpunit when using your plugin.
I've found a fix who makes it possible to run tests without any major changes.

A possibility to set the token to a false like value could also fix this issue.

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -125,7 +125,7 @@ class JWT
      */
     public function getToken()
     {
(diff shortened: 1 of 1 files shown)


tymondesigns:
Thanks but this will very tightly couple the package with laravel itself. No Laravel Facades have been used in this package for a reason. Dependency injection is preferred

PR #861, opened 2016-09-28
Use \Illuminate\Support\Str::random.

lucasmichot:
Use `\Illuminate\Support\Str::random` - `\Illuminate\Support\Str::quickRandom` is deprecated since `5.3`


tymondesigns:
👍

PR #867, opened 2016-10-05
Fix authorisation header being discarded by laravel

bertheyman:
By default, Laravel/Lumen are discarding the Authorisation Bearer token of the headers.
At this moment, this should be fixed by adding `RewriteRule ^ - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]`to the htaccess. As this is not in the docs, it it confusing for some people.

This pull request gets the header with php functions when the one from Laravel is empty and tries to fix the error that way. Hope this is useful. Fix based on a previous comment from @jeroenbourgois.

Any feedback is very welcome!

Update: will try and fix the tests.

### DIFF
--- a/src/Http/Parser/AuthHeaders.php
+++ b/src/Http/Parser/AuthHeaders.php
@@ -53,13 +53,13 @@ class AuthHeaders implements ParserContract
     {
         $header = $request->headers->get($this->header) ?: $this->fromAltHeaders($request);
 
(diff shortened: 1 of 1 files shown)


bertheyman:
Sorry, my bad!

I don't have much experience with tests, but I'll give it a go and try to look into it. If anyone would like to give me some tips or direction, feel free.

PR #875, opened 2016-10-19
- Fix: Refreshing Expired token fail with option refresh_ttl enabled

samymassoud:
When refresh_ttl Enabled , Refresh expired  token will fail with exception TokenExpiredException


samymassoud:
Please Cancel This Pull Request , i have managed to fix this by setting refresh flow to true before refresh.

Issue #877, opened 2016-10-23
Allow for using other field besides 'sub' such as 'uid'


jessica-mulein:
I was hoping to use firebase-compatible tokens for my app, and it seems they need 'sub' to be an email and instead want to use a 'uid' field.

https://firebase.google.com/docs/auth/server/create-custom-tokens

If I change my getJwtIdentifier() function to return the email address firebase wants, I can auth with firebase fine, but jwt-auth package is unhappy when it tries to use sub to get the ID back.

I might make a pull-request as an idea, but suspect you may have a more preferred way to allow this.

BrettTrotter:
Taking a quick stab at this, I first looked at making another function in JWTSubject that was getJwtIdentifierField() but then this could not easily be known during JWTAuth. Then I looked at adding a config item like JWT_IDENTIFIER_CLAIM_FIELD (defaulted to 'sub'), but config is not really easily accessible anywhere in JWT or JWTAuth where we'd need them. I was looking at passing the config data into the manager since the manager seems to be potentially around when we need it.. But these types of changes are going to be very subjective as to how you want to implement and I don't want to mess up your design ideals. I'll keep poking at something and I may submit a PR even if I'm ultimately sure you won't like the direction I took.

stueynet:
Watching. We use uuids for users.

PR #878, opened 2016-10-23
Issue #877: Accept other fields for id instead of subject

BrettTrotter:
I don't really like the duplication of the config() calls, but this was the general gist and a quick test with my login and with passing the token I get to firebase seems to work, but there may be more code needed to persist the 'sub' field anyway even if it's not the identifier field anymore?

Also FWIW, to work with firebase, I had to set JWT_ALGO to RS256, and set JWT_PUBLIC_KEY to a file path (\n's in .env don't seem to work) and a JWT_PRIVATE_KEY with a file:// path, and JWT_TTL to 30 minutes (3600 is the Firebase max).

Setting the new JWT_IDENTIFIER_CLAIM_FIELD env to 'uid' and my getJwtCusomClaims() to
```
	public function getJWTCustomClaims()
	{
		return [
			'iss' => '******@appspot.gserviceaccount.com',
			'sub' => '******@appspot.gserviceaccount.com',
			'aud' => 'https://identitytoolkit.googleapis.com/google.identity.identitytoolkit.v1.IdentityToolkit',
			'uid' => $this->getKey()
		];
	}
```

seems to get me up and going... I fully understand if this isn't the way you wish to implement.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -157,6 +157,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Identifier Claim Field
+    |--------------------------------------------------------------------------
+    |
+    | Specify the field to be used by JWTAuth for determining the unique
+    | identifier. Usually this is 'sub', but Firebase and other services
+    | may use another field, like 'uid'.
+    */
+
+    'identifier_claim_field' => env('JWT_IDENTIFIER_CLAIM_FIELD', 'sub'),
+
+    /*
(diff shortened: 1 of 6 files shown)


BrettTrotter:
Closing to redo in a single commit with fixed quotes, which coincidentally I copied from another location in your code, but there was some evaluated variables, thus the double quotes. Of course single is preferred where there is no evaluation.

PR #879, opened 2016-10-23
Issue #877: Accept other fields for id instead of subject

weeboflubber:
I was hoping to use firebase-compatible tokens for my app, and it seems they need 'sub' to be an email and instead want to use a 'uid' field.

https://firebase.google.com/docs/auth/server/create-custom-tokens

If I change my getJwtIdentifier() function to return the email address firebase wants, I can auth with firebase fine, but jwt-auth package is unhappy when it tries to use sub to get the ID back.

The code here seems to do the trick, but I don't really like the duplication of the config() calls. This was the general gist and a quick test with my login and with passing the token I get to firebase seems to work, but there may be more code needed to persist the 'sub' field anyway even if it's not the identifier field anymore?

Also FWIW, to work with firebase, I had to set JWT_ALGO to RS256, and set JWT_PUBLIC_KEY to a file path (\n's in .env don't seem to work) and a JWT_PRIVATE_KEY with a file:// path, and JWT_TTL to 30 minutes (3600 is the Firebase max).

my .env:

```
JWT_ALGO=RS256
JWT_IDENTIFIER_CLAIM_FIELD=uid
# A path or resource to your public key. E.g. 'file://path/to/public/key'
JWT_PUBLIC_KEY=file:///var/www/******/firebase.id_rsa_pub
# A path or resource to your private key. E.g. 'file://path/to/private/key'
JWT_PRIVATE_KEY=file:///var/www/******/firebase.id_rsa
# The passphrase for your private key. Can be null if none set.
#JWT_PASSPHRASE=
# max 3600 for Firebase
JWT_TTL=30
JWT_SECRET=******
JWT_BLACKLIST_ENABLED=true
```

my User.php:

```
    public function getJWTCustomClaims()
    {
        return [
            'iss' => '******@appspot.gserviceaccount.com',
            'sub' => '******@appspot.gserviceaccount.com',
            'aud' => 'https://identitytoolkit.googleapis.com/google.identity.identitytoolkit.v1.IdentityToolkit',
            'uid' => $this->getKey()
        ];
    }

    public function getJWTIdentifier()
    {
        return $this->getKey();
    }
```

seems to get me up and going... I fully understand if this isn't the way you wish to implement.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -157,6 +157,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Identifier Claim Field
+    |--------------------------------------------------------------------------
+    |
+    | Specify the field to be used by JWTAuth for determining the unique
+    | identifier. Usually this is 'sub', but Firebase and other services
+    | may use another field, like 'uid'.
+    */
+
+    'identifier_claim_field' => env('JWT_IDENTIFIER_CLAIM_FIELD', 'sub'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Blacklist Enabled
     |--------------------------------------------------------------------------
     |
--- a/src/Contracts/JWTSubject.php
+++ b/src/Contracts/JWTSubject.php
@@ -14,7 +14,7 @@ namespace Tymon\JWTAuth\Contracts;
 interface JWTSubject
 {
     /**
-     * Get the identifier that will be stored in the subject claim of the JWT.
+     * Get the identifier that will be stored in the identifier field claim of the JWT.
      *
      * @return mixed
      */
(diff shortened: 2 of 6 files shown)


BrettTrotter:
Well, locally and on my server (PHP 5.7) the config() calls work, not sure why the CI build fails, but I'm not sure what I need to do to fix for your checks.

BrettTrotter:
Darnit, github automatically pulled in a second commit needed for my project to use my repo after I had submitted the PR.. Now i'm not sure how to keep that commit but not have the PR take on any further commits.

rigobcastro:
@BrettTrotter Hi, I need verify ID tokens, with your pull-request is possible? Thanks!

BrettTrotter:
@rigobcastro Some conflicts have developed since i made the PR but it failed their checks because of the config() calls and i didn't have time to resolve. So the PR is not viable at this time but is perhaps a starting point for someone else?

PR #884, opened 2016-10-28
Fix for refresh token flow. Blacklisting token occurs before refresh token is generated invalidating token for refresh flow.

### LINKED ISSUE #808: RefreshToken is prematurely blacklisted
When using the 1.0.0-alpha.2 version of the RefreshToken with the "jwt.refresh" the token is first refreshed and then it is being checked for the next request. This result in the current token being blacklisted by the refresh() call and not proceeding to execute the request. 

```
public function handle($request, Closure $next)
    {
        $this->checkForToken($request);

        try {
            $token = $this->auth->parseToken()->refresh();
        } catch (JWTException $e) {
            throw new UnauthorizedHttpException('jwt-auth', $e->getMessage(), $e, $e->getCode());
        }

        $response = $next($request);

        // send the refreshed token back to the client
        $response->headers->set('Authorization', 'Bearer '.$token);

        return $response;
    }
```

should be:

```
public function handle($request, Closure $next)
    {
        $this->checkForToken($request);

        $response = $next($request);

        try {
            $token = $this->auth->parseToken()->refresh();
        } catch (JWTException $e) {
            throw new UnauthorizedHttpException('jwt-auth', $e->getMessage(), $e, $e->getCode());
        }

        // send the refreshed token back to the client
        $response->headers->set('Authorization', 'Bearer '.$token);

        return $response;
    }
```

mofodojodino:
Fixes #808

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -105,12 +105,12 @@ class Manager
      */
     public function refresh(Token $token, $forceForever = false)
     {
+        $payload = $this->setRefreshFlow()->decode($token, false);
+
         if ($this->blacklistEnabled) {
             // invalidate old token
             $this->invalidate($token, $forceForever);
         }
-
-        $payload = $this->setRefreshFlow()->decode($token, false);
 
         // persist the subject and issued at claims
         $claims = array_merge(


tymondesigns:
Since there were conflicts, I have implemented this myself via ae8c3f2bac1897a4441ebd9ed6fd77ed5eaa8a1e

Thank you!

mofodojodino:
Great, I wasn't sure if that was going to effect anything else. I'm just glad I could help find the issue for the project.

PR #895, opened 2016-11-03
Blacklisted tokens can be refreshed

BasConijn:
The blacklist is checked before refreshing the token.

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -112,12 +112,12 @@ class Manager
     {
         $this->setRefreshFlow();
 
+        $payload = $this->decode($token);
+        
         if ($this->blacklistEnabled) {
             // invalidate old token
             $this->invalidate($token, $forceForever);
         }
-
-        $payload = $this->decode($token, false);
 
         // assign the payload values as variables for use later
         extract($payload->toArray());
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -113,7 +113,7 @@ class Manager
         $this->setRefreshFlow();
 
         $payload = $this->decode($token);
-        
+
         if ($this->blacklistEnabled) {
             // invalidate old token
             $this->invalidate($token, $forceForever);


tymondesigns:
If a token is blacklisted then it should not be refreshable, otherwise there's no point in having a blacklist.

Or am I missing something?

BasConijn:
Yeah, that is exactly the problem.
An blacklisted token can be refreshed.
The pull request fixes that.

BasConijn:
I want to write an test to prove that an blacklisted token can be refreshed.
However I am running against an issue with mockery.
I have added this to the ManagerTest.php

`    /**
     * @test
     * @expectedException \Tymon\JWTAuth\Exceptions\TokenBlacklistedException
     */
    public function it_should_not_refresh_a_blacklisted_token()
    {
        $claims = [
            new Subject(1),
            new Issuer('http://example.com'),
            new Expiration($this->testNowTimestamp - 3600),
            new NotBefore($this->testNowTimestamp),
            new IssuedAt($this->testNowTimestamp),
            new JwtId('foo'),
        ];
        $collection = Collection::make($claims);

        $this->validator->shouldReceive('setRefreshFlow->check')->andReturn($collection);
        $payload = new Payload($collection, $this->validator);
        $token = new Token('foo.bar.baz');

        // Depending on the place in the code and parameters, this will or won't throw the exception
        $this->jwt->shouldReceive('decode')->with('foo.bar.baz')->andReturn($payload->toArray()); // ->andThrow(new TokenBlacklistedException());

        $this->factory->shouldReceive('setRefreshFlow')->with(true)->andReturn($this->factory);

        $this->blacklist->shouldReceive('has')->with($payload)->andReturn(true);

        $this->manager->refresh($token);
    }
`

tymondesigns:
This should now be resolved as per #1062

PR #911, opened 2016-11-18
Adding assert for the test of Namshi class

GabrielDeveloper:
I found one test that don't have assert to test behavior expected


tymondesigns:
Thanks!

PR #913, opened 2016-11-18
 Adding test to cover ReflectionException

GabrielDeveloper:
Covering the behavior when the algorithm configured is wrong


tymondesigns:
Thanks!, I might end up changing things slightly though to throw a package specific exception

PR #914, opened 2016-11-18
Reset claims collection when creating a payload.

wrseward:
This fixes an issue I've been having when using this library, specifically when writing tests with it.

The use case is that I create a token with some custom claims and then later want refresh the token and remove one of the claims.

For example if I use the following code
```php
<?php
/** @var JWTAuth $auth */
$auth = $this->app[JWTAuth::class];
$user = factory(User::class)->make();

$token = $auth->claims(['foo' => 'bar', 'baz' => 'qux'])
    ->fromUser($user);

dump($auth->setToken($token)->getPayload()->get());

$token = $auth->claims(['foo' => 'bar'])
    ->refresh();

dump($auth->setToken($token)->getPayload()->get());
```

I get the following output:
```
array:8 [
  "iss" => "http://:"
  "iat" => 1479501277
  "exp" => 1479504877
  "nbf" => 1479501277
  "jti" => "f48235a527b19245c2ade2937f2d4a04"
  "sub" => null
  "foo" => "bar"
  "baz" => "qux"
]
array:8 [
  "iss" => "http://:"
  "iat" => 1479501277
  "exp" => 1479504877
  "nbf" => 1479501277
  "jti" => "d2d9b3509c05474014f113b9eb37345a"
  "sub" => null
  "foo" => "bar"
  "baz" => "qux"
]
```

What I would expect is that the latter array to not have the `"baz"` key whatsoever, as I have specified a new set of custom claims.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -69,7 +69,10 @@ class Factory
      */
     public function make()
     {
-        return $this->withClaims($this->buildClaimsCollection());
+        $payload = $this->withClaims($this->buildClaimsCollection());
+        $this->claims = new Collection;
+
+        return $payload;
     }
 
     /**
--- a/tests/FactoryTest.php
+++ b/tests/FactoryTest.php
@@ -191,6 +191,34 @@ class FactoryTest extends AbstractTestCase
     }
 
     /** @test */
+    public function it_should_exclude_claims_from_previous_payloads()
+    {
+        $fooClaim = new Custom('foo', 'bar');
+        $barClaim = new Custom('baz', 'qux');
+
+        $this->claimFactory->shouldReceive('getTTL')->andReturn(60);
+        $this->claimFactory->shouldReceive('get')->with('foo', 'bar')->twice()->andReturn($fooClaim);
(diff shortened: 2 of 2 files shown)


tymondesigns:
Would you mind updating the tests too ?

wrseward:
Added a new test that should cover this case. Please let me know if you see anything else that I should correct. Thank you!

tymondesigns:
:+1:

dvdbrink:
@tymondesigns @wrseward This pull request causes getJWTCustomClaims claims to be removed after a token refresh. This is unexpected behaviour for me. I'm expecting basic user information (id, name and roles) in my token, even after a refresh, because the data could've been changed in the meantime.

tymondesigns:
Ahh yes you are right, OK I think I need to refactoring this on slightly to
make this behaviour optional

tymondesigns:
Ok I've now made this optional, as I think it's best to be "off" by default in this instance

See d4c2822b2eafafb6512e4ca0798a79ecbd8a2c4f

Thanks again @wrseward & @dvdbrink 

Any issues, let me know

eddiclin:
@tymondesigns 
How can I update the `iat` claim when refresh token?

PR #936, opened 2016-12-22
Blacklist sends NULL to Carbon

DarkGhostHunter:
If the key in `has` method isn't found, `Utils::isFuture` will send a NULL value to Carbon who will throw an exception. Can happen if storage doesn't find a blacklist key.


tymondesigns:
Thanks, this is indeed a valid edge case 👍

PR #969, opened 2017-01-24
Add Laravel 5.4 compatibility

rhwilr:
Laravel 5.4 removed the `share()` method as well the option to pass an array of parameters to `make()`.
This pull request fixes both of this in the JWTAuthServiceProvider

### DIFF
--- a/src/Providers/JWTAuthServiceProvider.php
+++ b/src/Providers/JWTAuthServiceProvider.php
@@ -119,8 +119,11 @@ class JWTAuthServiceProvider extends ServiceProvider
      */
     protected function registerUserProvider()
     {
-        $this->app['tymon.jwt.provider.user'] = $this->app->share(function ($app) {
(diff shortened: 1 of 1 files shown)


juanpscotto:
I have the same issue, could not use jwt with laravel 5.4 because of the share missing method

tymondesigns:
Thanks!

There's some other things I need to check on master branch before I do a release. But i'll do that asap

gustavobissolli:
@tymondesigns would be great release da beta version!

jackbarham:
I had the same issue. Had to revert back to 5.3. I'll subscribe to this thread for any updates :)

shahmir811:
After using the singleton function as committed by rhwilr (refactored to use singleton instead of share)  , I got the following error now:

Call to undefined method Illuminate\Database\Schema\MySqlBuilder::defaultStingLength()

Issue #972, opened 2017-01-24
LaravelServiceProvider not registering middleware properly?


isaackearl:
I'm on the dev-develop branch.  I just upgraded to Laravel 5.4 and now I get errors that say that jwt.auth doesn't exist.

```
ReflectionException in Container.php line 681:
Class jwt.auth does not exist
```

I have checked the LaravelServiceProvider and made sure I registered it in my app.php (this all works before I upgraded to 5.4)... this is the line that I'm assuming should create that middleware option:

```
$this->app['router']->middleware('jwt.auth', Authenticate::class);
```

I tried creating a middleware group like this:
```
 'auth:api' => [
            'throttle:200,1',
            'jwt.auth',
        ],
```

but I of course get the error... 

I have gotten it to work by using the middleware class directly instead of trying to use the name.

```
'auth:api' => [
            'throttle:200,1',
            Authenticate::class
        ],
```

So that is a workaround.. but I'm guessing this is not working as intended.

wuori:
Strange, the only thing I had to change was the provider in `app.php` (changed to `        Tymon\JWTAuth\Providers\LaravelServiceProvider::class`. 

This is probably known already but it's my first time working from the dev branch.

isaackearl:
@wuori Are you on laravel 5.4?  And you can use 'jwt.auth' in your middleware?  

I just retried installing everything just to make sure... and I still can't use jwt.auth middleware for the dvelop branch on 5.4.

juanpscotto:
jwt it's not 100% compatible with laravel 5.4 yet see https://github.com/tymondesigns/jwt-auth/pull/969

wuori:
@isaackearl Yes, just upgraded to 5.4 when I posted the comment. Verified version just in case, as it seemed a little odd to have it work without making other changes. Still using older middleware syntax in the route: 

``` php
Route::group(
    ['middleware' => ['auth']], function () {
```

isaackearl:
@wuori if you look at the LaravelServiceProvider class you'll see where the name 'jwt.auth' is registered for the Authenticate class... I was trying to use this name in order to use the middleware.  This worked fine before the upgrade.  Like this:

```
$this->app['router']->middleware('jwt.auth', Authenticate::class);
```

Because of this I've always been able to use jwt.auth as a middleware like this:
```
Route::group(
    ['middleware' => ['jwt.auth']], function () {
```

[...]

PR #976, opened 2017-01-25
fix LaravelServiceProvider for L 5.4

isaackearl:
In Laravel 5.4 there is now a aliasMiddleware method that does what the middleware function did before.  I do an if statement check so the middleware will be compatible with previous versions of Laravel 5 as well.  This is a fix for issue #972


tymondesigns:
Awesome thanks.. Not had the chance to play around with 5.4 yet

PR #979, opened 2017-01-27
Suggest HTTP-status-code 498 like ArcGIS

willgopublic:
I would like to suggest to use http-status-code 498, like ArcGIS seems to do for tokens.
«498 Invalid Token (Esri)
Returned by ArcGIS for Server. A code of 498 indicates an expired or otherwise invalid token.[69]
» ~ https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#cite_ref-arcgis_70-0 , http://help.arcgis.com/en/arcgisserver/10.0/apis/soap/index.htm#Using_token_authentication.htm .

### DIFF
--- a/src/Exceptions/TokenExpiredException.php
+++ b/src/Exceptions/TokenExpiredException.php
@@ -11,6 +11,8 @@
 
 namespace Tymon\JWTAuth\Exceptions;
 
+use Exception;
+
 class TokenExpiredException extends JWTException
 {
     /**
--- a/src/Exceptions/TokenExpiredException.php
+++ b/src/Exceptions/TokenExpiredException.php
@@ -13,5 +13,16 @@ namespace Tymon\JWTAuth\Exceptions;
 
 class TokenExpiredException extends JWTException
 {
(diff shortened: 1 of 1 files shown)


src/Exceptions/TokenExpiredException.php
```diff
@@ -13,5 +13,15 @@
...
 {
-    //
+    /**
+     * @param  string  $message default: 'Token Expired'
+     * @param  int  $code default: 498
+     * @param  \Exception|null  $previous default: null
+     *
+     * @return void
+     */
+    public function __construct($message = 'Token Expired', $code = 498, Exception $previous = null)
```
lucasmichot: `\Exception` should be imported.

lucasmichot:
Tests?

lucasmichot:
@willgopublic, I would avoid using unofficial HTTP codes...

tymondesigns:
Yea I don't really wan't to add a non-standard http status code into the package. You are free to respond to your clients however you wish though

PR #1010, opened 2017-02-18
Throw if user no longer exists in authentication middleware

salaman:
Currently, in the 1.0 branch, the auth middleware will not throw an exception if the JWT token is valid, but the user in the `sub` claim no longer exists. This functionality existed in pre-1.0 versions.

In previous versions, this threw a 404, however I believe a 401 is more appropriate.

It may also be worth invalidating the token in such cases, but I didn't add it into this PR.

### DIFF
--- a/src/Http/Middleware/BaseMiddleware.php
+++ b/src/Http/Middleware/BaseMiddleware.php
@@ -66,9 +66,13 @@ abstract class BaseMiddleware
         $this->checkForToken($request);
 
         try {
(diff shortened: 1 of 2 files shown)


tymondesigns:
sorry for the delay. Thanks!

d3radicated:
Does it get the user from the default guard? What if the user may come from different guards (e.g. API server for two back-end apps)?

PR #1018, opened 2017-02-23
Let the first letter capitalize

sc0Vu:
After read https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html, I think maybe we should make the authorization header first letter capitalize.


tymondesigns:
This actually makes no difference since we use `stripos` to parse the headers which is case insensitive (https://github.com/tymondesigns/jwt-auth/blob/develop/src/Http/Parser/AuthHeaders.php#L56)

Issue #1024, opened 2017-02-28
Should status code be 400 or 401 if token is not provided


mehrdaad:
The title is self explanatory, If ``Authorization`` header is absent what would be appropriate status code?  
I see if u use ``api.auth`` middleware from Dingo it will be 401 while jwt will be 400.

andrewmclagan:
Agreed, i was about to raise the same issue before I found this.

- `400` denotes a mal-formed HTTP request.
- `401` denotes the the request was unauthorised 

Logic would dictate a request failing to provide a token is `401` not authorised rather then `400` badly formed. 

Clients accessing a RESTful resource expect response codes to explain correctly what has happened. In this case authentication failure. The client can thus handle a typical `401`.

Now clients are typically a frontend website, mobile app or other services. Responding to the root cause which is "missing auth" is simpler "Hey i got 401 back, there fore i will display the unauthorized dialog" rather then "Hey I got 400 back, i will display the 'something went wrong' dialog".

The offical specs says `401` is 

> specifically for use when authentication is required ... or has not yet been provided.

## Official REST specs

**400 Bad Request**
The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, too large size, invalid request message framing, or deceptive request routing) [...]

**401 Unauthorized**
Similar to 403 Forbidden, but specifically for use when authentication is required and has failed or has not yet been provided. [...]

mehrdaad:
@andrewmclagan thanks, but this seems like a bug rather than a decision.  
Issuing ``400`` status code for not providing token wrong at least for me.  
Do you agree with me too?

andrewmclagan:
Sure i could do this.

mehrdaad:
@andrewmclagan appreciated.

andrewmclagan:
@mehrdaad Just finding time to fix the issues with the PR now

tymondesigns:
merged :)

manuel-84:
I'm having this issue again with latest v

dennisvang:
>Official REST specs

(from [comment above])

@andrewmclagan Could you provide a link to these "official" REST specs? 

I only know https://ics.uci.edu/~fielding/pubs/dissertation/top.htm

[comment above]: https://github.com/tymondesigns/jwt-auth/issues/1024#issuecomment-284622542

PR #1028, opened 2017-03-08
Correct status code for missing auth tokens

andrewmclagan:
This was discussed in https://github.com/tymondesigns/jwt-auth/issues/1024

`400` represents a mal-formed request.

`401` is specifically for use when authentication is required or has not yet been provided.

### DIFF
--- a/src/Http/Middleware/BaseMiddleware.php
+++ b/src/Http/Middleware/BaseMiddleware.php
@@ -48,7 +48,7 @@ abstract class BaseMiddleware
     public function checkForToken(Request $request)
     {
         if (! $this->auth->parser()->setRequest($request)->hasToken()) {
-            throw new BadRequestHttpException('Token not provided');
+            throw new UnauthorizedHttpException('Token not provided');
         }
     }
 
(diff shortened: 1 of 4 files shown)


andrewmclagan:
@mehrdaad in reference to issue https://github.com/tymondesigns/jwt-auth/issues/1024#issuecomment-285476512 This PR is failing due to what seems a phpunit dependancy issue with php 5.7.x. Should i bother fixing this or leave the PR as is?

mehrdaad:
@andrewmclagan thanks, just a quick note:  
I think in files: ``AuthenticateAndRenewTest.php``, ``RefreshTokenTest.php`` and ``AuthenticateTest.php`` test method name should be ``it_should_throw_an_unauthorized_exception_if_token_not_provided`` instead of ``it_should_throw_a_bad_request_exception_if_token_not_provided``.  
probably you forgot to refactor them. 😃

tymondesigns:
Ok I'm sold on this now.. it has cropped up a few times before.

One small thing.. could you change the test method names to reflect the change also?

cgarnier:
Yes, it should be a 401. Maybe you can merge that ? :)

cgarnier:
In version 0.5.10, it throws a 404 when the user is not found. I dont know if it is the case in this one.
But, it should be a 401 too. It s true that the user is not found but the errror code is for the resource we are getting not for the auth purpose.

tymondesigns:
I'll merge and change the methods

mehrdaad:
@tymondesigns apparently it was forgotten to refactor test method names so ``it_should_throw_a_bad_request_exception_if_token_not_provided`` should be renamed to ``it_should_throw_an_unauthorized_http_exception_if_token_not_provided``.  
right?

andrewmclagan:
@tymondesigns apologies I was busy last week at work. Thanks for taking care of that.

PR #1039, opened 2017-03-17
Extract middleware aliases

lucasmichot:
Extract middleware aliases
And determine only once if `aliasMiddleware` function exists


tymondesigns:
👍 good catch

PR #1040, opened 2017-03-17
Fix PHPUnit.

lucasmichot:
Fix PHPUnit.


lucasmichot:
@tymondesigns, I believe this PR could be merged quickly so that all others PRs can be re-built

tymondesigns:
Cheers!

PR #1043, opened 2017-03-19
Some tests fixes.

lucasmichot:
Some tests fixes.


tests/Validators/PayloadValidatorTest.php
```diff
@@ -145,7 +145,7 @@ public function it_should_throw_an_exception_when_providing_an_invalid_expiry()
 
         $collection = Collection::make($claims);
 
-        $this->validator->check($payload);
+        $this->validator->check($collection);
```
tymondesigns: I wonder if phpunit uses instanceof to check if an exception is thrown during a method call, thus surpressing a blatant error like this

PR #1056, opened 2017-03-24
Remove useless constructor.

lucasmichot:
Remove useless constructor.

### DIFF
--- a/src/Claims/Collection.php
+++ b/src/Claims/Collection.php
@@ -17,18 +17,6 @@ use Illuminate\Support\Collection as IlluminateCollection;
 class Collection extends IlluminateCollection
 {
     /**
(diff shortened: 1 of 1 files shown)


tymondesigns:
This was necessary because the behaviour was changed in laravel 5.2, 

(5.1) https://github.com/laravel/framework/blob/5.1/src/Illuminate/Support/Collection.php#L33-L36
(5.2+) https://github.com/laravel/framework/blob/5.2/src/Illuminate/Support/Collection.php#L34-L37

Because of these methods from memory
https://github.com/tymondesigns/jwt-auth/blob/develop/src/Claims/Collection.php#L96-L120

I agree though it is not ideal!

Issue #1060, opened 2017-03-25
After Blacklisting a token I can still refresh it.


isaackearl:
I discovered what I think of as a Critical issue with the development branch.  If I blacklist a token (logout), I can refresh it and then use it again without re-authenticating...

Here is the series of things I'm doing to reproduce.

login 

hit a protected route (jwt.auth) -> let's me in

logout (blacklist/invalidate the token)

hit the protected route again (401 unauthorized your token has been blacklisted message)

hit the refresh endpoint (jwt.refresh) -> 200 OK

hit the protected route again (jwt.auth) -> let's me in!

I've already checked I have blacklist enabled, and I don't have a grace period on blacklisted tokens.

tymondesigns:
Just looked at this again, and you're correct. The regression was introduced [here](https://github.com/tymondesigns/jwt-auth/commit/6c9e05c10ba812481f2d83073d44a4b634547c46)

Going to push a fix now and do some more investigation

tymondesigns:
Could you pull in the latest and check that it fixes the problem for you please?

isaackearl:
@tymondesigns pulling it in now, I will report back soon.

isaackearl:
@tymondesigns Ya that seems to have fixed this.  Thanks!  I was really scratching my head trying to debug it.

tymondesigns:
Awesome thank you for raising 👍

isaackearl:
@tymondesigns I may open another issue, but wanted to tell you here first.  I think fixing that issue, causes another one.

I am now unable to use the refresh middleware to actually refresh the token (even during the refresh_ttl because it checks the blacklist.  and I get the error about the token being on the blacklist.  I'll see if I can find an easy fix and make a pull request however I'm not so sure taking a brief look, so you may have to handle it yourself.  It would be a shame however if these 2 issues keep regressing back and forth, so it's a good idea to keep them both in mind.

francisrod01:
I had the same problem, but in this case in `dev-develop` branch.
It is was solved there also. :+1: 
Thanks!

PR #1062, opened 2017-03-25
invalidate the old token AFTER building the claim

isaackearl:
After the fix for issue #1060 the refresh middleware stopped working because it would check the blacklist and then fail because the token is already invalidated.  If you move the invalidation til after the "buildRefreshClaims" method then you don't have this problem.

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -113,13 +113,13 @@ class Manager
     {
         $this->setRefreshFlow();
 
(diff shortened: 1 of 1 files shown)


tymondesigns:
haha literally just been testing this EXACT fix!

tymondesigns:
Thanks! 🎉

isaackearl:
haha nice!

PR #1064, opened 2017-03-25
change to a token invalid exception

isaackearl:
After the fix for #1060 and the subsequent pull request, I found another issue I hope you will consider.

As it stands right now, the refresh functionality is hard to use.  

When I navigate to an endpoint with an expired token I get a TokenExpiredException, this lets me know that I need to refresh the token.  I then hit the refresh endpoint and as it stands currently will either 1) refresh the token, or 2) give me another TokenExpiredException with the message "Token has expired and can no longer be refreshed".  It should return something else like an invalid token exception etc.  

As it currently stands I can't tell the difference between an expired token that can be refreshed, and an expired token that cannot be refreshed.  The only way I can tell the difference is by looking at the error message.  This seems hacky and improper since the message could change with any push to the repo.

### DIFF
--- a/src/Claims/IssuedAt.php
+++ b/src/Claims/IssuedAt.php
@@ -12,7 +12,6 @@
 namespace Tymon\JWTAuth\Claims;
 
 use Tymon\JWTAuth\Exceptions\InvalidClaimException;
-use Tymon\JWTAuth\Exceptions\TokenExpiredException;
 use Tymon\JWTAuth\Exceptions\TokenInvalidException;
 
 class IssuedAt extends Claim
@@ -52,7 +51,7 @@ class IssuedAt extends Claim
     public function validateRefresh($refreshTTL)
     {
         if ($this->isPast($this->getValue() + $refreshTTL * 60)) {
-            throw new TokenExpiredException('Token has expired and can no longer be refreshed');
+            throw new TokenInvalidException('Token has expired and can no longer be refreshed');
         }
     }
 }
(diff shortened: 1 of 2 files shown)


isaackearl:
I understand this change is more then a little fix, but I hope you'll consider it.  As it stands right now it shouldn't change the functionality of your middleware, because they all catch the exception and throw an UnauthorizedHttpException, this just makes it possible for me to tell the difference between a refreshable token and one that is not.

tymondesigns:
Good point.. although I would prefer it if we had another exception that extends `TokenExpiredException ` maybe something like `RefreshTokenExpiredException` or something similar. Then backwards compatibility would be maintained since it would still be an instance of `TokenExpiredException`

isaackearl:
That is much smarter.  I’ll make another pull request.

isaackearl:
@tymondesigns Ok I think that should do it.

isaackearl:
hey @tymondesigns 

So it ends up that I was thinking to hard about a problem, and I was able to do what I wanted without this pull request simply by creating my own middleware.  I'm going to go ahead and just close this because I don't think you should merge it anymore.

tymondesigns:
ok yea.. I always advise people to create their own middleware's, but in this instance I kind of like the idea of adding this extra exception, since it is non-breaking and adds the ability to distinguish between the possible outcomes...

isaackearl:
Well when I was implementing it, I thought it would be much better overall if this same token was included in the Expiration class.  

If you are going to have 2 exceptions that can help you know the difference between a refreshable expired token and a non refreshable expired token then you would need to modify this function as well in the Expiration class:

```
/**
     * {@inheritdoc}
     */
    public function validatePayload()
    {
        if ($this->isPast($this->getValue())) {
            throw new TokenExpiredException('Token has expired');
        }
    }
```

I think this change would be really good but only if it was consistent across the package.  I might sit down and try and implement it at some point if you like the idea.

PR #1069, opened 2017-03-28
Lots of small fix

lucasmichot:
Lots of small fix


src/Providers/Auth/Illuminate.php
```diff
@@ -17,11 +17,15 @@
 class Illuminate implements Auth
 {
     /**
+     * teh authentication guard.
```
tymondesigns: Typo :)

lucasmichot: Fixed !

tymondesigns:
👍 Thanks for taking the time to do these little improvements

PR #1074, opened 2017-03-29
More tests fixes

lucasmichot:
Some more test fixes
And add AbstractMiddlewareTest

### DIFF
--- a/tests/AbstractTestCase.php
+++ b/tests/AbstractTestCase.php
@@ -25,9 +25,7 @@ abstract class AbstractTestCase extends TestCase
     {
         parent::setUp();
 
-        $now = Carbon::now();
-
-        Carbon::setTestNow($now);
+        Carbon::setTestNow($now = Carbon::now());
         $this->testNowTimestamp = $now->getTimestamp();
     }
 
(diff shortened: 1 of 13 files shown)


tymondesigns:
👍

lucasmichot:
Looks like a stable release soon :-)

tymondesigns:
Yes it does!

tymondesigns:
Already planning 1.1 release.. dropping laravel 5.1 and removing all the auth stuff that isn't guard related.

Also considering php 7 only.. what do you think?

lucasmichot:
I would go for droping not guard related stuff 

But forcin php>=7 would lock force to L>=5.5
Any specific reason for that ?

tymondesigns:
None other than, because I want to play around more with some of the newer features. Type hint all the things etc 😀

But I think I should be a little bit responsible and resist the temptation.. could be a shit storm of php 5.* Peeps

lucasmichot:
Ah ah ah !
As long as previous version plays nice with PHP=<7.0 and L=<5.5, I believe you can make it

lucasmichot:
Tough choice

tymondesigns:
I'll see how things play out, once 1.0.0 is out the door

PR #1088, opened 2017-04-04
JTI needs to be better random

gm-ghanover:


tymondesigns:
Thanks i don't mind tagging a release for this

PR #1154, opened 2017-05-13
Refactor the command.

lucasmichot:
Refactor the command.


tymondesigns:
Cheers.. might have another go at adding tests for this

lucasmichot:
Yep would make sense !

lucasmichot:
In fact, after having thought a bit about it - I think I would just remove this command as this is obvious that a token needs to be set.

I would only raise an exception when the size of the token is < 32 characters.

tymondesigns:
@lucasmichot That's a fair point.. I'll have a think about it

Issue #860, opened 2016-09-28
Laravel 5.3 Setup "Guide"


mtpultz:
Okay, I've got a bit of time, and need to set up a new project with Laravel 5.3 and JWTGuard.  @tymondesigns I hope you don't mind this being in your issues similar to my other "guide" (#513), and if so I can try and find a different spot, but I don't write blogs or tutorials.

This is just a post of the steps to get the [JWTGuard](https://github.com/tymondesigns/jwt-auth/blob/develop/src/JWTGuard.php) in place for an API using Laravel 5.3.x to possibly help with starting the documentation of this feature.  Anything from #513 that is the same is still included so you don't need to browse through it unless I missed something.  I've set this up twice now, and I found myself that it is just easier to do a clean install, and port.  So this assumes you are doing a clean install, and currently are using version v5.3.11 of Laravel .

Also, jwt-auth 1.0 hasn't been released yet so this also assumes that you are pulling in the development branch `"tymon/jwt-auth": "dev-develop",` using composer.

Using JWTGuard with Laravel 5.3.x
=====

1) `config/app.php` - add `Tymon\JWTAuth\Providers\LaravelServiceProvider::class` to providers
2) In your terminal publish the config file: `php artisan vendor:publish --provider="Tymon\JWTAuth\Providers\LaravelServiceProvider"`
3) In your terminal generate the secret: `php artisan jwt:secret` provided you have JWT_SECRET in your .env file
4) `config/auth.php` - set the default guard to `api`, and change the `api` driver to `jwt`

```
'defaults' => [
    'guard' => 'api',
    'passwords' => 'users',
],
```

```
'guards' => [
    ...

    'api' => [
        'driver' => 'jwt',
        'provider' => 'users',
    ],
],
```

5) `/routes/api.php` - add a few basic authentication routes

```
Route::group([
    'prefix' => 'api'
], function () {

    Route::post('login', 'Auth\LoginController@login');
    
    Route::group([
        'prefix' => 'restricted',
        'middleware' => 'auth:api',
    ], function () {

        // Authentication Routes...
        Route::get('logout', 'Auth\LoginController@logout');

        Route::get('/test', function () {
            return 'authenticated';
        });
    });
});
```

6) `/app/LoginController.php`

[...]

PR #1167, opened 2017-05-27
Added Support for Multi Auth (Multiple Guards)

cappuc:
This add support to Laravel Multi-Auth (Multiple Guards) and fix the problem with permissions of the tokens reported here #860

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -236,6 +236,7 @@ class JWT
     {
         return array_merge(
             ['sub' => $subject->getJWTIdentifier()],
+            ['guard' => $subject->getJWTGuard()],
             $this->customClaims, // custom claims from inline setter
             $subject->getJWTCustomClaims() // custom claims from JWTSubject method
         );
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -236,10 +236,34 @@ class JWT
     {
         return array_merge(
             ['sub' => $subject->getJWTIdentifier()],
-            ['prv' => get_class($subject)],
+            ['prv' => $this->hashProvider($subject)],
             $this->customClaims, // custom claims from inline setter
             $subject->getJWTCustomClaims() // custom claims from JWTSubject method
         );
+    }
+
+    /**
+     * Hash the provider and return it.
+     *
+     * @param  string|object  $provider
+     *
+     * @return string
+     */
+    protected function hashProvider($provider)
+    {
+        return hash('sha256', is_object($provider) ? get_class($provider) : $provider);
+    }
+
+    /**
+     * Check if the provider matches the one saved in the token.
+     *
+     * @param  string|object  $provider
+     *
+     * @return bool
+     */
+    public function checkProvider($provider)
+    {
+        return $this->hashProvider($provider) === $this->payload()->get('prv');
     }
 
     /**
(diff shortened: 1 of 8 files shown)


tymondesigns:
What problem does this solve exactly?

cappuc:
If you have multiple Models like Customer and Admin, the admin token can be used to authenticate the customer and vice-versa, which is incorrect.

I have added a claim to the token that save the Model class name and when JWTGuard try to retrieve the user it checks if the guard provider correspond to the provider saved into the token.

src/JWTGuard.php
```diff
@@ -74,9 +74,14 @@ public function user()
 
         if ($this->jwt->getToken() && $this->jwt->check()) {
             $id = $this->jwt->payload()->get('sub');
```
lucasmichot: It's useless to resolve `$id` outside `if ($this->provider->getModel() == $provider) {`

src/JWTGuard.php
```diff
@@ -74,9 +74,14 @@ public function user()
...
+            $provider = $this->jwt->payload()->get('prv');
 
-            return $this->user = $this->provider->retrieveById($id);
+            if ($this->provider->getModel() == $provider) {
+                $this->user = $this->provider->retrieveById($id);
```
lucasmichot: No need to go further, you can just `return $this->user = $this->provider->retrieveById($id);`
as the final `return $this->user;` (`null`) is useless

src/JWTGuard.php
```diff
@@ -73,9 +73,13 @@ public function user()
         }
 
         if ($this->jwt->getToken() && $this->jwt->check()) {
```
lucasmichot: Maybe an early return would make more sense?
```php
if ($this->jwt->getToken() === false || $this->jwt->check() == false) {
    return;
}

if ($this->provider->getModel() === $this->jwt->payload()->get('prv')) {
    $id = $this->jwt->payload()->get('sub');

    return $this->user = $this->provider->retrieveById($id);
}
```

lucasmichot: Maybe an early return would make more sense?
```php
if ($this->jwt->getToken() === false || $this->jwt->check() == false) {
    return;
}

if ($this->provider->getModel() === $this->jwt->payload()->get('prv')) {
    $id = $this->jwt->payload()->get('sub');

    return $this->user = $this->provider->retrieveById($id);
}
```

src/JWT.php
```diff
@@ -236,6 +236,7 @@ protected function getClaimsArray(JWTSubject $subject)
     {
         return array_merge(
             ['sub' => $subject->getJWTIdentifier()],
+            ['prv' => get_class($subject)],
```
tymondesigns: I think it might be cool to change this to be an md5.. as a cheap means of obfuscation. e.g.

`md5(get_class($subject))`

I would also actually pull this out into it's own method

tymondesigns: Or even better use `hash('sha256', get_class($subject));`

cappuc: Yes, it's a good idea to obfuscate the class. I'll make a commit to add this.

tymondesigns:
Thanks for this!

I think I will tweak it a little post merge and maybe make this optional, so people can still keep their tokens smaller if they don't need this behaviour

PR #1190, opened 2017-06-07
Allow null prv claim.

crynobone:
This allows JWT token created before tymondesigns/jwt-auth#1167 can be
used until TTL is expired. Would be useful instead of logout everyone since the token failed
checkProvider().

Signed-off-by: crynobone <crynobone@gmail.com>


tymondesigns:
good catch Thanks!

PR #1207, opened 2017-06-14
Use SHA1 instead of MD5.

lucasmichot:
Using `md5` to encode the `prv` claim (which is readable clear) seems a bit weak as there are many ways to reverse it - I believe that no one wants to provide too much details about which model is used for authentication.

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -267,7 +267,7 @@ class JWT
      */
     protected function hashProvider($provider)
     {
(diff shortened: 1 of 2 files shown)


tymondesigns:
I'm aware of the issues with md5, The reason I went back on this is because I wanted to keep the length as small as possible, and didn't put as much importance in someone finding out the model path if they wanted to. But I guess your right...

PR #1208, opened 2017-06-15
Use get method instead of query

OmarMakled:
When we use query method token should pass as query string and in many cases we post token with form inputs thats why even if you passed valid token you got token_not_provided


tymondesigns:
Thanks for this, but I'm no longer accepting non-critical fixes to the master branch

urukalo:
is there any option to go around this?
i have problem with fcn IE, and only solution is to have token in POST body. This PR allow this, very CRITICAL, feature for me

Issue #1221, opened 2017-06-22
Claims array should override standard claims array


miscbits:
In this function here: 

![image](https://user-images.githubusercontent.com/8864479/27442083-47c0a26e-573d-11e7-8a1a-9a2746075417.png)

We are seeing that we can update our jwt custom claims by using the custom claims function, but the array merge is in such an order that the custom array has its values overwritten by the default array. 

For our use case, we have a situation where we want to update a user's token to reflect some changes, but their token has a default value (in the database) so we need to be able to overwrite this without changing the user's information in the database. Basically 

```
// App\User
public function getJWTCustomClaims() {
    return ['value1'=>'foo'];
}

// App\Http\Controller\AuthController

public function updateToken() {
    $customClaims = ['value1'=>'bar'];
    return JWTAuth::customClaims($customClaims)->fromUser(Auth::user())->getPayload();
    // returns ['value1'=>'foo'] when it should return ['value1'=>'bar']
}
```

If we just switch lines 239 and 240 in Jwt.php, then the problem is resolved.

PR #1222, opened 2017-06-22
changed order of array merge in JWT

miscbits:
to allow inline to overwrite default custom claim

Addresses issue in #1221


tymondesigns:
I actually agree with this, Thanks!

Issue #1219, opened 2017-06-20
problem with lumen 5.4.27


olivecashlez:
Hi,

I've been updated my lumen composer, to v5.4.27, before this I had v5.4.19 and It was good, I got no error. The error is:

> (1/1) JWTException
The token could not be parsed from the request
in JWT.php (line 176)
at JWT->parseToken()

I'm using v1.0.0-beta.3 
 "tymon/jwt-auth": "^1.0@dev",

Thank you.

chrisschaetzlein:
Same problem here

michielyat:
It seems to be that the parser does not know the request variable anymore. 

Edit: I forgot to mention that this solution is located in the JWTGuard.php file.

Before the following code snippet

```
        if ($this->jwt->getToken() && $this->jwt->check()) {
            $id = $this->jwt->payload()->get('sub');

            return $this->user = $this->provider->retrieveById($id);
        }
```

Add the following code line:

`$this->jwt->parser()->setRequest($this->request);`

By this the parser that is looking for the request headers can now read the headers again. Somehow when the parser is initialized the current request object is not send to the parser at runtime (it keeps working with a default empty request).

Hopefully the creator of the package has a solution for this matter. Meanwhile this solution works and I'm able to test JWT with Lumen. If this is a correct solution? I'm not sure because I'm still learning laravel / lumen etc :-) ...

abrarshareef:
I have the same problem.

Adding the line to JWTGuard.php worked:
`$this->jwt->parser()->setRequest($this->request);`

Please fix.

daniel-mcassey-tmsuk:
Yup, seems Lumen has lost access to the request in the function "registerTokenParser()" in "AbstractServiceProvider.php" the $app['request'] is empty.

Further testing shows that this is broken in 5.4.2 but works in 5.4.1, so another fix is to mark the laravel/lumen-framework in composer.json to stick to 5.4.1 then do composer update

tymondesigns:
closed by #1243

PR #1223, opened 2017-06-22
#1219 Fix for missing request in JWTGuard

michielpeeters:
Fixed JWTGuard when it tries to parse the authorization header

### DIFF
--- a/src/JWTGuard.php
+++ b/src/JWTGuard.php
@@ -72,6 +72,8 @@ class JWTGuard implements Guard
             return $this->user;
         }
 
+        $this->jwt->parser()->setRequest($this->request);
+        
         if ($this->jwt->getToken() &&
             ($payload = $this->jwt->check(true)) &&
             $this->jwt->checkProvider($this->provider->getModel())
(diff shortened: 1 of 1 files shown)


src/JWTGuard.php
```diff
@@ -72,6 +72,8 @@ public function user()
             return $this->user;
         }
 
+        $this->jwt->parser()->setRequest($this->request);
+        
         if ($this->jwt->getToken() &&
```
tymondesigns: It would be better to change it to 
```php
if ($this->jwt->setRequest($this->request)->getToken() && // etc
```

tymondesigns (changes requested):

src/JWTGuard.php
michielpeeters: Agree, I have it changed in a new commit :-)

tymondesigns:
Just need to update tests and fix style and then I'm happy to merge this

tymondesigns:
@michielpeeters Are you comfortable tweaking the tests for this?

tymondesigns:
#1243 now implements this.

Thanks for the effort @michielpeeters

PR #1225, opened 2017-06-24
Blacklist - Simplifying the `has` method

KennedyTedesco:
`empty()` evaluates true when the value is null.

### DIFF
--- a/src/Blacklist.php
+++ b/src/Blacklist.php
@@ -128,13 +128,8 @@ class Blacklist
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


tymondesigns:
Looks good.. would you mind adding a couple of test cases for this? If not then I can test it out myself

KennedyTedesco:
Added some tests, feel free to change them if you want to. 😊

tymondesigns:
👍

PR #1229, opened 2017-06-28
on refresh update the token

kevupton:
This is to prevent the blacklist exception on payload retrieval.

(If the token is refresh at the start of the request, then there will be a blacklist exception on getPayload)

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -97,7 +97,7 @@ class JWT
 
         $this->token = $this->manager->customClaims($this->getCustomClaims())
                              ->refresh($this->token, $forceForever, $resetClaims);
-        
+
         return $this->token->get();
     }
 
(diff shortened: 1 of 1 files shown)


hopeseekr:
It looks like it doesn't change anything??

tymondesigns:
The logical flow is exactly the same here so not sure what your issue is ?

kevupton:
I was having an issue where if I have some middleware to refresh a token, then later on request the payload during the controller of my application, it would get a token blacklist exception. As it was using the older token, instead of the newly refreshed token. By setting `$this->token`, it will use the new token in the `getPayload` request. 

The refresh is happening at the beggining of the request.

tymondesigns:
This has already been covered with `JWT_BLACKLIST_GRACE_PERIOD` in the config. Simply set this to something higher than 0

kevupton:
That is right, however that setting would allow for multiple requests at once, using the same token. This is not the intended functionality that I am after.

PR #1243, opened 2017-07-06
Fix JWT Guard and add tests #1219

DanielMcAssey:
This should fix the missing request mentioned in #1219

### DIFF
--- a/src/JWTGuard.php
+++ b/src/JWTGuard.php
@@ -72,7 +72,7 @@ class JWTGuard implements Guard
             return $this->user;
         }
 
-        if ($this->jwt->getToken() &&
+        if ($this->jwt->setRequest($this->request)->getToken() &&
             ($payload = $this->jwt->check(true)) &&
             $this->jwt->checkProvider($this->provider->getModel())
         ) {
--- a/tests/JWTGuardTest.php
+++ b/tests/JWTGuardTest.php
@@ -128,6 +128,7 @@ class JWTGuardTest extends AbstractTestCase
      */
     public function it_should_return_null_if_an_invalid_token_is_provided()
     {
+        $this->jwt->shouldReceive('setRequest')->andReturn($this->jwt);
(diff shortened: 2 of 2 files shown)


tymondesigns:
Thanks!

gemins:
I dont understand. Not see any change in my vendor :( I used "composer update", "composer required", clear cache, but always return the version 1.0.0-beta-3 (in composer.json say: "tymon/jwt-auth": "^1.0@dev") and the file JWTGuard.php in line 75 say:

```
        if ($this->jwt->getToken() && $this->jwt->check()) {
            $id = $this->jwt->payload()->get('sub');

            return $this->user = $this->provider->retrieveById($id);
        }

```
but i see in this git this:

```
if ($this->jwt->setRequest($this->request)->getToken() &&
($payload = $this->jwt->check(true)) &&
$this->jwt->checkProvider($this->provider->getModel())
) {
return $this->user = $this->provider->retrieveById($payload['sub']);
 }

```

What i'm doing wrong?

wiretab:
Was struggling with this issue where calling parseToken() would throw the "token could not be parsed from request" error. Was using composer's ^1.0@dev version, still had the issue. So I cloned the git branch manually and replaced it.

Even with the fix shown here in JWTGuard.php, if you var-dump the request object's headers in Parser.php's parseToken() method, it shows a null array. I traced that functionality back to the top and it seems the request just isn't being set somewhere along the line.

My fix was so manually set the request as follows in my controller:

`    public function Home(Request $request)
    {
        $this->jwt->setRequest($request);
        $user = $this->jwt->parseToken()->authenticate();
        var_dump($user);
    }
`

For me, this fixes the issue. I'm using Lumen v5.4. 

Edit: this is from some half-awake debugging and the application I'm creating is still pretty patchwork-y, so this may be some bad advice. Use at your own peril!

Edit 2: it seems (at least for me) the $app['request'] being used to create a new Parser() object in AbstractServiceProvider is a default (empty) request object.

PR #1247, opened 2017-07-11
Factorize blacklist test.

lucasmichot:
Factorize blacklist test.


tymondesigns:
Cool! never used data providers

PR #1248, opened 2017-07-11
No useless else.

lucasmichot:
- No useless else in StyleCI
- Prefer null comparison

---

This is working fine : https://styleci.io/analyses/8jooOG

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,6 +1,8 @@
 preset: laravel
 
 enabled:
+  - is_null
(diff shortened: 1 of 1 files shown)


tymondesigns:
`no_useless_else` removes ternary operators?

lucasmichot:
My bad @tymondesigns this was a test commit - it has been removed

tymondesigns:
@lucasmichot ahh makes sense! Thanks dude

Issue #1242, opened 2017-07-05
Get "Unauthorized" in all router with middleware


gemins:
I'm using Tymon/Jwt-auth : 1.0@dev and Lumen 5.4.

After update my lumen from 5.4.19 to 5.4.27 was recive this message in all routes when is protected by middleware with JWT.

I'm using native AuthServiceProvider:
```
$this->app['auth']->viaRequest('api', function ($request){
            if($request->input('dni'))
                return User::where('dni', $request->input('dni'))->first();
            else if ($request->input('username'))
                return User::where('dni', $request->input('username'))->first();
            else if ($request->input('email'))
                return User::where('dni', $request->input('email'))->first();
}
```

In bootstrap/app.php
```
$app->routeMiddleware([
    'auth' => App\Http\Middleware\Authenticate::class
]);
$app->register(App\Providers\AppServiceProvider::class);
$app->register(App\Providers\AuthServiceProvider::class);
//$app->register(App\Providers\EventServiceProvider::class);
$app->register(Tymon\JWTAuth\Providers\LumenServiceProvider::class);
```

Before update my lumen, i was login with the system normaly and haven't problem. Now can login and get the Token, but when try to enter in a route with middleware i get the response: "Unauthorized"

Thanks you.

sandbox-services:
confirmed: your lumen guide fails unless you set your lumen version to 5.4.1 (5.4.2->5.4.6) all return Unauthorized regardless if you're sending the jwt token via Authorize headers or token in query string

tymondesigns:
closed by #1243

Issue #1241, opened 2017-07-05
Request headers empty in vendor/tymon/jwt-auth/src/Http/Parser/Parser.php


jesperbjerke:
Trying to use the parseToken() in a controller method but it keeps saying it couldn't retrieve the token from the request. Digging into that, the request in my controller method has headers:

```php
public function refresh(Request $request)
{
        echo '<pre>';
            var_dump($request->header());
        echo '<pre>';
        die();
}
```
produces
```
/app/Http/Controllers/AuthController.php:118:
array (size=12)
  'accept-language' => 
    array (size=1)
      0 => string 'en-US,en;q=0.8' (length=14)
  'accept-encoding' => 
    array (size=1)
      0 => string 'gzip, deflate' (length=13)
  'accept' => 
    array (size=1)
      0 => string '*/*' (length=3)
  'content-type' => 
    array (size=1)
      0 => string 'text/plain;charset=UTF-8' (length=24)
  'user-agent' => 
    array (size=1)
      0 => string 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.115 Safari/537.36' (length=115)
  'origin' => 
    array (size=1)
      0 => string 'chrome-extension://fhbjgbiflinjbdggehcddcbncdddomop' (length=51)
  'cache-control' => 
    array (size=1)
      0 => string 'no-cache' (length=8)
  'postman-token' => 
    array (size=1)
      0 => string 'XXXXXX' (length=36)
  'authorization' => 
    array (size=1)
      0 => string 'Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJodHRwOi8vYXBpLnRpbmdyLmFwcC9hdXRoL2xvZ2luIiwiaWF0IjoxNDk5MTg3NjIwLCJleHAiOjE0OTkxODc2ODAsIm5iZiI6MTQ5OTE4NzYyMCwianRpIjoiSE5oa0lHbFg4R1ZWVndRcyIsInN1YiI6MTJ9.E3snWGvCc7I-iuf8udg9sLpgiqruBLbqJpoG7ATn83' (length=255)
  'content-length' => 
    array (size=1)
      0 => string '305' (length=3)
  'connection' => 
    array (size=1)
      0 => string 'keep-alive' (length=10)
  'host' => 
    array (size=1)
[...]

PR #1252, opened 2017-07-12
Update JWTGuard.php

mikield:
Setting the request to JWT instance before getting Token and checking it.
In cause of lumen the request was allways a new installce (empty request) and not the send.

I think that this is a fast fix, and this sollution can be fixed better.

Should fix: #1242 #1241 #1219


tymondesigns:
Thanks but this has already been done #1243

mikield:
@tymondesigns at witch version is it fixed? Had this problem using 1.0.0-beta.3 version.

PR #1255, opened 2017-07-16
Add handle method for Secret Generator Command

hughsaffar:
Laravel 5.5 is looking for the handle method instead of fire. "handle" method would call "fire" method.

### DIFF
--- a/src/Console/JWTGenerateSecretCommand.php
+++ b/src/Console/JWTGenerateSecretCommand.php
@@ -33,6 +33,16 @@ class JWTGenerateSecretCommand extends Command
     protected $description = 'Set the JWTAuth secret key used to sign the tokens';
 
     /**
+     * This method deligates to Laravel 5.5 would look for 
+     * "handle" method instead of fire
+     * @return void
+     */
+    public function handle()
+    {
+        $this->fire();
+    }
+
+    /**
      * Execute the console command.
      *
      * @return void


tymondesigns:
I'll allow the spelling mistake 😉 Thanks

vesper8:
@tymondesigns I think handle() has been the de facto method name to use since Laravel 5.0

there's really no reason to keep fire() like this PR did. And it actually broke a script of mine that I use to auto-fix this for packages that haven't changed their fire->handle yet which resulted in a redeclared handle() method

it's unnecessary and no one else is doing it this way.. you should just rename your fire() to handle() and get rid of this little backward compatibility snippet IMHO

tymondesigns:
@vesper8 the fire method was only renamed to handle about a month ago (laravel 5.5 only). 

So the fire method is needed until I decide to drop support for everything less than 5.5 (probably the next major release)

vesper8:
@tymondesigns, apologies if I was unclear, what I meant to point out is that

handle() has been supported ever since Laravel 5.0

In 5.5 fire() is now completely deprecated. But a handle() call would work on all versions greater than 4.2. Thus there is no reason to keep the fire() call unless you wanted to support Laravel 4

https://github.com/laravel/framework/blob/4.2/src/Illuminate/Console/Command.php#L112
https://github.com/laravel/framework/blob/5.0/src/Illuminate/Console/Command.php#L113
https://github.com/laravel/framework/blob/5.5/src/Illuminate/Console/Command.php#L180

tymondesigns:
Ah ok now I understand what you mean. Yea we should remove the `fire` method

jonathaniqueda:
why this PR has not accepted yet?

Chintan7027:
Any solution for this ? I am facing same problem

hughsaffar:
The PR is already merged into the "develop" branch.
You can try:
`
composer require tymon/jwt-auth:dev-develop --prefer-source
`

leandroruel:
used the command above, after this, i get this:
Class 'Tymon\JWTAuth\Providers\JWTAuthServiceProvider' not found

zekker6:
@leandroruel Seems like it is called `Tymon\JWTAuth\Providers\LaravelServiceProvider::class` now

danielemontecchi:
I added `Tymon\JWTAuth\Providers\LaravelServiceProvider::class,` at list of providers in `config/app.php`.
For generate JWTAuth secret key, use: `php artisan jwt:secret`

lahin31:
@akajack after using `php artisan jwt:secret` my JWTAuth secret didn't update automatically.

iHmD:
@lahin31 updated in the .env file at the bottom ;)

hoanglong2205:
is `aliases` part stay the same ?

hiteshkan123:
@tymondesigns Thanks, renaming the function worked for me :+1:

skcin7:
So... what's the recommended solution for this?

hiteshkan123:
@skcin7 Same, I think renaming the fire function to handle will work..

leandroruel:
the fix is only in the develop branch?

spfroc:
thanks it works.

MaksymGrom:
@skcin7  You can use my solution:
extend your fixes like in this file
https://github.com/StanislavGrom/blog.loc/tree/master/app/Console/JWTFixes
And add rewrite via Laravel IoC
https://github.com/StanislavGrom/blog.loc/blob/master/app/Providers/AppServiceProvider.php (16-19 lines)
Now Laravel will resolve class creation, and return JWTFixes instead original class.

ErcilioMarques:
This is because the command class adheres to the Laravel 4 command interface, which uses a fire() method instead of handle(). As a quick and dirty fix you can add the following to Tymon\JWTAuth\Commands\JWTGenerateCommand:
``

public function handle()
{
    $this->fire();
}

roynes:
@MaksymGrom Thanks! This works, Overriding it is much safer while waiting for the package to be updated.

I hope @tymondesigns's PR would be merged to the **0.5**

jesus143:
> The PR is already merged into the "develop" branch.
> You can try:
> `composer require tymon/jwt-auth:dev-develop --prefer-source`

This answer saved my hour. Thank you so much.

aeharake:
Go to this file **Tymon\JWTAuth\Commands\JWTGenerateCommand**

add the method:

```
 public function handle(){
        $this->fire();
 }

```
Some IDEs will ask you if you're sure you want to edit the file, just say Yes and Save.

PR #1275, opened 2017-08-01
Update composer.json

ragingdave:
Add L5.5 to enable usage.


ziazon:
Laravel 5.5 is due out any day now, can we get this merged in so we can start using it? 👍

PR #1288, opened 2017-08-20
Simplify illuminate package versioning

BafS:
To avoid putting `||` for each new version.


jubairsaidi:
👍

tymondesigns:
Thanks for this, but since Laravel does not really do semver, and minor releases often have many breaking changes, it feels better to be explicit for now

jubairsaidi:
I guess fork it is for now :(

PR #1291, opened 2017-08-22
Use stdClass.

lucasmichot:
Use `stdClass`.


tymondesigns:
what's the difference? 🤔 😃

PR #1318, opened 2017-09-19
Backward compatibility

axotion:
We can decorate it instead of renaming function


tymondesigns:
Thanks but the code already works going back to laravel 5.0 so no need to add this

PR #1328, opened 2017-09-27
to string token

layodemetrio:


tymondesigns:
Thanks but i'm not really sure this is a worthwhile addition, can you explain the use case?

PR #1339, opened 2017-10-10
fix dead links

dv336699:


tymondesigns:
The docs are actually hosted on [readthedocs](http://jwt-auth.readthedocs.io/en/docs/laravel-installation/) so these links are fine, Thanks for the PR anyways

PR #1354, opened 2017-10-21
Exceptions caught and raised accordingly

jampack:
The exceptions can be handled in exception handler class whenever the middleware is brought to use before 'auth' middleware. Useful in Lumen 5.5 as the normal response from auth middleware is a HTTP 401 without any description.


akkhan20:
Style issues, closing.

PR #1357, opened 2017-10-23
Remove some require dependencies

ralbear:
The required Illuminate/* packages are now a part of laravel/framework


tymondesigns:
Thanks, but the point is to depend on _exactly_ what we need rather than the whole framework,

PR #1359, opened 2017-10-24
Add handle alias to JWTGenerateCommand for Laravel 5.5

patcarter883:


tymondesigns:
This is already addressed in the new version `1.*`, and there are no plans to support laravel 5.5 via the `0.5.*` releases. Thanks

PR #1363, opened 2017-10-29
feat: decrypt cookies during parsing

akiver:
Hi, since Laravel encrypt cookies by default, currently if you don't tell Laravel to not encrypt your cookie token (https://laravel.com/docs/5.5/responses#cookies-and-encryption), the parsing will fail.

This PR add an option to decrypt or not cookies which is by default enabled to follow Laravel behavior.

Note: it's a breaking change for those who currently added their cookie in the except middleware array.

Thank you

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -202,6 +202,23 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Cookies encryption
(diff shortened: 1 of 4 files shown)


tymondesigns:
This is great, thank you!

PR #1373, opened 2017-11-06
Adjusted regex to allow comma-seperated tokens in the Authorization h…

dhixson:
The current regex does not find the prefixed token if more than one item is in the authorization header as it captures everything including the comma to separate. This regex change will find the prefixed token even if it is not the only token in the Authorization header.

This feature would be useful for developers making a system that requires both an API key(for application level authentication and authorization) and a Token(for user level authentication and authorization).

Thank you for the great authentication tool!

### DIFF
--- a/src/Http/Parser/AuthHeaders.php
+++ b/src/Http/Parser/AuthHeaders.php
@@ -53,7 +53,7 @@ class AuthHeaders implements ParserContract
     {
         $header = $request->headers->get($this->header) ?: $this->fromAltHeaders($request);
 
-        if ($header && preg_match('/'.$this->prefix.'\s*(.*)\b/i', $header, $matches)) {
+        if ($header && preg_match('/'.$this->prefix.'\s*(\S+)\b/i', $header, $matches)) {
             return $matches[1];
         }
     }


Kyslik:
I wonder:

1) Is there a standard for this?
2) should JWT-auth support:
 a) `Authorization: Bearer f1RsTtoKen, Bearer s30ndtoKeN`
 b) `Authorization: Bearer f1RsTtoKen, s30ndtoKeN`

dhixson:
Good questions. Here are my answers.

1. The closest I can find to a standard is [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.2.2) which states.

> A sender MUST NOT generate multiple header fields with the same field name in a message unless either the entire field value for that header field is defined as a comma-separated list [i.e., #(values)] or the header field is a well-known exception (as noted below).

> A recipient MAY combine multiple header fields with the same field name into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field value to the combined field value in order, separated by a comma. The order in which header fields with the same field name are received is therefore significant to the interpretation of the combined field value; a proxy MUST NOT change the order of these field values when forwarding a message.

2. By this standard I think the support should be for `Bearer s30ndtoKeN, Bearer2 s31ndtoKeN`, but not for duplicates under the same key.

tymondesigns:
Interesting... I will look into this some more myself

I'm happy to merge since it is a non breaking change

Thanks!

PR #1383, opened 2017-11-12
Imwhocodes patch get token()

imwhocodes:
Make getToken() try to parse request only the first time

In the case of absent/invalid token avoid to try to parse the request every time getToken() is called by parsing it only the first time and saving the result of it.

The return behavior is unchanged

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -154,11 +154,11 @@ class JWT
      */
     public function getToken()
     {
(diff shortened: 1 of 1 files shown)


tymondesigns:
The behaviour effectively remains the same here, since you set the token to false and then check if it is null. I have added the check in myself anyways since it's such a small thing. Cheers!

PR #1394, opened 2017-11-20
Check if email was sent in login request

rafaelrenanpacheco:
The AuthController from quick start allows to login if the email is not sent and if the password matches the first user from the database.

For example, the following post do `/login` will try to login with the first user found in database:
```json
{
	"password": "{{ password  }}"
}
```

Which generates the following SQL:
```
select * from user limit 1
```

This PR ensures that the email was sent in the request. If not, then 401.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -135,7 +135,7 @@ class AuthController extends Controller
     {
         $credentials = request(['email', 'password']);
 
-        if (! $token = $this->guard()->attempt($credentials)) {
+        if (! isset($credentials['email']) || ! $token = $this->guard()->attempt($credentials)) {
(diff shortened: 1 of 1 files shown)


tymondesigns:
That behaviour is very bizarre.. and in fact I am wondering how that is even possible, given that the EloquentUserProvider [handles that side of things](https://github.com/tymondesigns/jwt-auth/blob/develop/src/JWTGuard.php#L121)

In any case, this PR doesn't really solve the problem (if there even is one)

rafaelrenanpacheco:
Hi @tymondesigns!

I agree there is a deep problem around this, but this PR is the same changes I did to my code to ensure that the login does't run without the email input.

Best regards,
Rafael Pacheco.

PR #1402, opened 2017-11-30
Cookie decryption now set to false by default

jonothedev:
Cookie encryption is enabled by default in some frameworks such as Laravel.

Enabling encryption by default will cause invalid payload exceptions where double encryption is attempted.

Therefore i have set this to false to get round such issues.


tymondesigns:
I need to really take a better look at the implications of this and do some testing myself, hang fire

tymondesigns:
Yea I think this is a better default 👍

PR #1407, opened 2017-12-04
Rename `fire` method to `handle`

jerguslejko:
This PR renames the `JWTGenerateCommand::fire` to `handle` as it is required by Laravel 5.5.

Please note that this DOES NOT break backwards compatibility as Laravel versions prior to v5.5 supported both `fire` and `handle` methods.

### DIFF
--- a/src/Commands/JWTGenerateCommand.php
+++ b/src/Commands/JWTGenerateCommand.php
@@ -36,7 +36,7 @@ class JWTGenerateCommand extends Command
      *
      * @return void
      */
-    public function fire()
+    public function handle()
     {
         $key = $this->getRandomKey();
 


dinho19sp:
It is not necessary to rename the fire() method, you can create the handle() {} method and call the fire method on the handle() .. ex: public function **handle() {$ this->fire (); }**.

Please 
fix everyone bugs, for we uses frequently the jwt-auth

jerguslejko:
hey @dinho19sp, sure you can do that but you will end up with two methods instead of one? The change proposed in this PR in **non-breaking**, therefore safe to implement.

tymondesigns:
Thanks but no plans to patch `0.5` with this. If you want to support Laravel 5.5 then you would need to use the `1.0.*` release

PR #1408, opened 2017-12-04
Update JWTGenerateCommand.php

dinho19sp:
Compatibility with laravel 5.5,

Add method handle() in the file /src/Commands/WTGenerateCommand.php

for resolve the exception:

> In BoundMethod.php line 135:
> 
> Method Tymon\JWTAuth\Commands\JWTGenerateCommand::handle() does not exist


dongilbert:
+1 This should be used over #1407

dongilbert (approved):

tymondesigns:
See my comment on #1407

PR #1411, opened 2017-12-08
Use assertCount

carusogabriel:


tests/PayloadTest.php
```diff
@@ -199,7 +199,7 @@ public function it_should_get_the_object_as_json()
     public function it_should_count_the_claims()
     {
         $this->assertSame(6, $this->payload->count());
-        $this->assertSame(6, count($this->payload));
+        $this->assertCount(6, $this->payload);
         $this->assertCount(6, $this->payload);
```
tymondesigns: Thanks but this is redundant, as we perform the same check underneath

carusogabriel: I guess the new one will give us a better error message, in case it fails, but is up to you!

tymondesigns: You mean if `assertCount` comes before `assertSame` ?

carusogabriel: We are tests the `same` thing twice 🤔
Let’s keep `assertCount`? Because under the hood, it’s doing `assertSame($expected, count($actual))`

tymondesigns: ok well in that case let's remove `$this->assertSame(6, count($this->payload));` you want to update the PR?

carusogabriel: Did it. In case you are interesting, [here is the `PHPUnit` class that test `assertCount`](https://github.com/sebastianbergmann/phpunit/blob/6.5/src/Framework/Constraint/Count.php#L55), and proves what I'm talking :sweat_smile:

tymondesigns:
:+1:

PR #1412, opened 2017-12-08
Update composer.json

g0shed:
Wrong namespace for auto package discovery


tymondesigns:
This references the old path to the service provider, there is no official Laravel 5.5 support for `v0.5.*`. That is over on the `1.0.0` release

Issue #1425, opened 2017-12-20
jwt:generate broken!


paparyku:
**php artisan jwt:generate** - is broken!
```

php artisan jwt:generate

In BoundMethod.php line 135:

  Method Tymon\JWTAuth\Commands\JWTGenerateCommand::handle() does not exist
```

helpceo:
me too

rdarcy1:
This is because the command class adheres to the Laravel 4 command interface, which uses a `fire()` method instead of `handle()`. As a quick and dirty fix you can add the following to `Tymon\JWTAuth\Commands\JWTGenerateCommand`:

```php
// JWTGenerateCommand.php

public function handle()
{
    $this->fire();
}
```

Alternatively, the issue is fixed in the development version so you could pull in that version and use the [new docs](https://jwt-auth.readthedocs.io/en/docs/laravel-installation/). No idea how stable the dev version is at the moment.

1e4:
Since we are doing dirty hacks you can also just change the method `fire()` to `handle()` within `\vendor\tymon\jwt-auth\src\Commands\JWTGenerateCommand.php`

tymondesigns:
Yes I'm aware of this.. version `0.5.*` doesn't have full support for Laravel 5.5. You would need to use the newer release

PR #1466, opened 2018-01-28
Fix jwt:generate failing

devfunsho:
Fixed the issue where php artisan jwt:generate was failing in Laravel 5.5


rdarcy1:
This was discussed in #1425, doesn't look like @tymondesigns wants to fix in the older version.

jasperboyd (approved):

PR #1467, opened 2018-01-29
Add JWT::flushToken() method

rdarcy1:
Enables the token to be unset and ensures it will not be automatically parsed from the request etc. later on in the code.

I have a use-case where a user makes a request to an endpoint with a JWT, which is parsed and then they're logged in with jwt-auth and a new token issued. Currently I cannot unset the first token because it's just re-parsed from the request during `JWTGuard::attempt()` so I get an exception `You must unsign before making changes`.

Not sure if this is the best way to solve, open to feedback!

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -51,6 +51,13 @@ class JWT
     protected $lockSubject = true;
 
     /**
+     * Indicates if the parser should attempt to parse a new token when $this->token is null.
+     *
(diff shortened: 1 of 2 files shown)


Messhias (approved):

Messhias:
@rdarcy1 only @tymondesigns can accept merge requests?

PR #1470, opened 2018-02-01
Fixed command to generate a key not working laravel version less than 5.4.17

brucewu16899:
Fixed command to generate a key not working in laravel version less than 5.4.17


tymondesigns:
Good catch! 👍

PR #1493, opened 2018-02-22
Adding configurable header name and method.

CuCTeMeH:
In case someone uses HTTP_BASIC AUTH under development and needs JWT implementation the headers are not working well together. This commit fixes it as it makes the headers and header_method configurable.


tymondesigns:
Sorry but I'm not accepting any new features on master, develop is where it's at

PR #1501, opened 2018-03-02
Add Parser chain key

Except10n:


tymondesigns:
what's the purpose of this?

Except10n:
such a call

```
app('tymon.jwt.parser')->getChain()[ParserCookies::class]->setKey('xxxx');
```

@tymondesigns

tymondesigns:
Thanks for this, but going to address it a slightly different way

PR #1505, opened 2018-03-05
Added JWTProvider facade

linaspasv:
Hello,

I was looking forward to use same library for emails validations too but could not find a clean way to easly generate tokens outside of auth. After some search and thinking I came up with this PR, and I hope many are going to find specific uses of this in their projects!

Use-case example:

```php
use JWTProvider;

$payload = ['email' => 'customer@domain.com', 'referred_by' => 12];
$token = JWTProvider::encode($payload);

// we e-mail a confirmation link with the generated token to "customer@domain.com"

```

```php
use JWTProvider;

try {
    $payload = JWTProvider::decode($token);
} catch (Exception $e) {

}

// after token being validated, we could finalize the registration process

```

Thank you for a great work you did with this library! Keep up the good work.

### DIFF
--- a/src/Facades/JWTProvider.php
+++ b/src/Facades/JWTProvider.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of jwt-auth.
+ *
+ * (c) Sean Tymon <tymon148@gmail.com>
+ *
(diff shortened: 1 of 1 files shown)


tymondesigns:
Ok yea, this could be useful for others too

Thanks!

PR #1521, opened 2018-03-13
make lock_subject config available for JWTAuth

xcaptain:
I like use `JWTAuth::fromUser($u);` to generate jwt token, this pull request makes `lock_subject` config makes effect.


xcaptain:
Also I have a comment on the previous change, see https://github.com/tymondesigns/jwt-auth/commit/7c1d58e294b255c4afa31c121b11b947e8699cbe#comments

tymondesigns:
:+1:

PR #1535, opened 2018-03-22
Correct provider name in installation process

richin13:


tymondesigns:
The docs here are for the newer 1.0.0 version so would be incorrect

PR #1545, opened 2018-04-06
Make JWTAuth respect the "lock_subject" setting in the config file

4026:
I noticed that tokens generated via the JWTAuth facade still had the "prv" claim attached to them, regardless of the "lock_subject" setting in the configuration file. I've altered the instantiation of JWTAuth in the service provider so that it calls lockSubject() immediately on the instance (matching the procedure used for JWT earlier in the file).


tymondesigns:
:+1:

PR #1572, opened 2018-05-16
Wrong casing

ebsp:
Class imported as `Tymon\JWTAuth\JWTGuard` but referenced as `JwtGuard`, which create errors when running in HHVM.


tymondesigns:
Thanks!

PR #1576, opened 2018-05-23
Update the defaults as no required claims

l4u:
According to https://tools.ietf.org/html/rfc7519#section-4.1 , the Registered Claim Names are mandatory. Note that this PR has not been tested.


tymondesigns:
These claims are indeed optional as per the spec, which is why they are configurable, but I think the sensible default is to make them required by default

Issue #1137, opened 2017-05-01
error in reset claims


ashutoshSce:
in file Factory.php
```
public function make($resetClaims = false)
    {
        if ($resetClaims) {
            $this->emptyClaims();
        }

        $payload = $this->withClaims($this->buildClaimsCollection());

        return $payload;
    }
```

empty claims before creating payload based on $resetClaims flag

tymondesigns:
not sure what your point is?

ashutoshSce:
in file Factory.php

Current code
```
public function make($resetClaims = false)
    {
        $payload = $this->withClaims($this->buildClaimsCollection());

         if ($resetClaims) {
            $this->emptyClaims();
         }

        return $payload;
    }
```

Expected Code
```
public function make($resetClaims = false)
    {
        if ($resetClaims) {
            $this->emptyClaims();
        }

        $payload = $this->withClaims($this->buildClaimsCollection());

        return $payload;
    }
```

emptyClaims before creating payload...

This is my point bro :)

ashutoshSce:
Plz have a look.

empty claims before creating payload if resetClaims flag is true.

eddiclin:
Is this issue fixed yet?
I can't reset the claims when refresh the token, even thought the parameter `$resetClaims` is true.
See `Tymon\JWTAuth\Manager::refresh`

PR #1586, opened 2018-06-06
make refresh claim reset optional

eddiclin:
Resolve these issues:
#914 #1137 

Improve test cases:
https://github.com/tymondesigns/jwt-auth/commit/d4c2822b2eafafb6512e4ca0798a79ecbd8a2c4f#r29269188

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 build
 coverage.xml
+.idea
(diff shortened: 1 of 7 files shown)


eddiclin:
@tymondesigns 
In my opinion, `Tymon\JWTAuth\Manager::$refreshFlow` and `Tymon\JWTAuth\Factory::$claims` need some optimization.
Because they are singleton, if set once, it will affect future operations, such as `refresh`, `check`, unless the developer reset it manually.
Any ideas?

eddiclin:
reverted error

PR #1588, opened 2018-06-06
Decode token string only one time.

eddiclin:
The `Token` and `Payload` objects have a high cohesion. Make the `Token` object has a `Payload` property, we just need decode it only one time and use it in many places.


tymondesigns:
I have toyed around with a few approaches to achieve this myself, but not settled on any that I like so far.
Ultimately I don't think this is how I would like it to work exactly, but I'm going to leave this open for now. Thanks for getting the ball rolling and reminding me about this!

tymondesigns:
Implemented differently on the 2.0 branch

PR #1592, opened 2018-06-07
Fixed public function handle() { $this->fire(); } bug

rannieo:
Hi @tymondesigns,

I've add the missing method in `src\Commands\JWTGenerateCommand`, the said method was `public function handle() { $this->fire(); } `.

Version affected: 0.5.*

Hope you check it. :) Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor
 composer.lock
 build
+.idea/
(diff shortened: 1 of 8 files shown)


tymondesigns:
Thanks but v0.5.* is not intended to be used with Laravel 5.5 and above https://github.com/tymondesigns/jwt-auth/wiki#compatibility

PR #1593, opened 2018-06-07
Fix security bug in `JWTAuth::authenticate`

eddiclin:
`JWTAuth::authenticate` must to check the user's subject model. If not, we can use user's token to get admin's ability.

# Reproduce
Create a Laravel Project with `App\User` and `App\Admin` models. Config `config/auth.php` file, as below, for demonstration purposes the default guard is "admin".
``` php
return [
    'defaults' => [
        'guard' => 'admin',
    ],
    'guards' => [
        'api' => [
            'driver' => 'jwt',
            'provider' => 'users',
        ],
        'admin' => [
            'driver' => 'jwt',
            'provider' => 'admins',
        ],
    ],
    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],
        'admins' => [
            'driver' => 'eloquent',
            'model' => App\Admin::class,
        ],
    ]
];
```
Run code in `php artisan tinker`
``` php
$userToken = auth('api')->tokenById(1);
app('tymon.jwt.auth')->setToken($userToken)->authenticate();  // Error: Get the admin's info
```

# Fix
Check user's subject model.

After fix:
``` php
$userToken = auth('api')->tokenById(1);
app('tymon.jwt.auth')->setToken($userToken)->authenticate();  // return false
```

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -285,11 +285,13 @@ class JWT
      */
     public function checkSubjectModel($model)
     {
-        if (($prv = $this->payload()->get('prv')) === null) {
-            return true;
+        $prv = $this->payload()->get('prv');
+
+        if ($this->lockSubject || null !== $prv) {
+            return $this->hashSubjectModel($model) === $prv;
         }
 
-        return $this->hashSubjectModel($model) === $prv;
+        return true;
     }
 
     /**
--- a/src/JWTAuth.php
+++ b/src/JWTAuth.php
@@ -61,13 +61,9 @@ class JWTAuth extends JWT
      */
     public function authenticate()
     {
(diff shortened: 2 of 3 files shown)


tymondesigns:
I will pull this down and test locally when I get a minute

tymondesigns:
Had some time to look at this again, see my comment here https://github.com/tymondesigns/jwt-auth/issues/1769#issuecomment-726699543

This is essentially a misunderstanding of the middleware contained within this package. The guard implementation should only be used with laravel Authenticate middleware or similar.

This middleware is all going to be removed in the next major release

Thanks

PR #1597, opened 2018-06-08
feat(TokenKeyName): Set a configurable token key name

ElRochito:


config/config.php
```diff
@@ -27,6 +27,17 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | JWT Token Key Name
+    |--------------------------------------------------------------------------
+    |
+    | You can set a custom token key name in your .env file
+    |
+    */
+
+    'token_key_name' => env('JWT_TOKEN_KEY_NAME', 'token'),
```
tymondesigns: I don't think this needs to be configurable at the env level

tymondesigns (changes requested):

config/config.php
bastien-phi: Maybe just `'token_key_name' => 'token',` should be better ?

ElRochito (commented):
I've changed config/config.php to set a token name directly in config/jwt.php

ElRochito (commented):
Configurable in config.jwt.php instead of env file

ElRochito (commented):
I've updated my code

tymondesigns:
Thank you for your contribution but looking at this again, I don't think it's necessary.

PR #1604, opened 2018-06-19
Adding macroable capabilities to the Guard

tykus:
Laravel's Guard classes are macroable; adding the same capability to the JWTGuard gives to possibility to _extend_ the class at runtime.


tymondesigns:
Cool!

PR #1616, opened 2018-07-05
Don't reset blacklist expiration time

mattmcdev:
Currently, a blacklisted token can be used for an infinite amount of time when using the blacklist grace period, so long as a request is made at least once every x seconds (where x is the value of `blacklist_grace_period`).

This change ensures that the reset time on a blacklisted token is not reset every time a request is made.

### DIFF
--- a/src/Blacklist.php
+++ b/src/Blacklist.php
@@ -71,6 +71,11 @@ class Blacklist
             return $this->addForever($payload);
         }
 
+        // if we have already added this token to the blacklist
+        if (! empty($this->storage->get($this->getKey($payload)))) {
+            return true;
+        }
+
         $this->storage->add(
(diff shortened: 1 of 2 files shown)


tymondesigns:
Sorry for the _very_ late response, I think this is valid point and I'm happy to add it if you would be so kind to fix the tests? If not then I will pull this down and adjust myself when I have time

mattmcdev:
I've updated the tests to `allows` the `get` call, but this was introduced in mockery 1.0.0 so fails for PHP < 5.6.

Does PHP < 5.6 still need to be supported? Or do you know a better way to let the tests pass?

tymondesigns:
@mattmcdonald-uk I have pulled down your PR and fixed the issues here - see #1775

PR #1619, opened 2018-07-10
Refresh custom claims from the subject when a token is refreshed

### LINKED ISSUE #891: Update custom claims with refresh token
I use refresh token middleware for every request I make to my server and it works fine. I ran into a problem when i wanted to update custom claims (user first name, last name and some more). I update user and then I want updated token (with updated user claims) to be present in response header. So how can I set this updated token to be as response header? I tried to set it manually but the refresh token middleware overrides it after it handles request and uses old claims from request (not updated ones).

### LINKED ISSUE #1388: custom claim not working as expected
Hello,

based on documentation, the following code is supposed to work:
`
$credentials = $request->only('email', 'password');
$customClaims = ['sCustomercode' => '1234', 'sActivesessionID' => 'abcd'];
if (!$token = app('auth')->guard($this->guard)->attempt($credentials, $customClaims)) {
    return response()->json(['error' => 'invalid_credentials'], 401);
}
`

but when token is decoded, I have no custom properties.

The only way for now is:
`
$token = JWTAuth::fromUser($user, $customClaims);
$payload = JWTFactory::sCustomercode('1234')->sActivesessionID('1234')->sVerification('abcd')->make();
$token = JWTAuth::encode($payload)->get();
`
It is more intuitive to provide an array in attempt(), is it a bug or the feature changed without refreshing the documentation?

Thanks.

njbarrett:
Fixes #891 

Currently when you refresh a token, it does not regenerate the custom claims data. If for example data in my `getJWTCustomClaims` is dynamic the only way the token gets updated is with a new login.
This change will recreate the claims array properly for refreshed tokens as well.

(Apologies if this is the wrong way to do it, happy to discuss other solutions)

### DIFF
--- a/src/JWT.php
+++ b/src/JWT.php
@@ -102,7 +102,7 @@ class JWT
     {
         $this->requireToken();
 
-        return $this->manager->customClaims($this->getCustomClaims())
+        return $this->manager->customClaims($this->getClaimsArray($this->authenticate()))
                              ->refresh($this->token, $forceForever, $resetClaims)
                              ->get();
     }


mattmcdev:
This change breaks the refresh middleware flow as it requires the token not to be expired when refreshing.

If you first call `$this->manager->setRefreshFlow()` it appears to fix it.

A further problem is that `authenticate` may return `false` which can't be passed to `getClaimsArray`

njbarrett:
Hey @mattmcdonald-uk , thanks for your comment.
So are you saying if I add `$this->manager->setRefreshFlow()` instead of my code, or in addition to my code, I can fix this issue properly?

mattmcdev:
I had to make a couple of changes to get this fix working for me.

Adding a call to `setRefreshFlow` makes sure that tokens which can be renewed are used.

Adding a test on the call to `authenticate` ensures it doesn't fail if the user has disappeared and false is returned.

https://github.com/mattmcdonald-uk/jwt-auth/commit/8111ae719478258426d0a2ea2d6bf8650cdda114

mjsarfatti:
I get an error when I try to implement this:

```json
{
    "message": "Method [authenticate] does not exist.",
    "exception": "BadMethodCallException",
    "file": "[...]vendor/tymon/jwt-auth/src/JWT.php",
    "line": 425,
    [...]
```

This is because the `refresh()` method is declared in JWT.php, but `authenticate()` is in JWTAuth.php.
Am I missing something?

mattvb91:
@mjsarfatti stuck on exactly this issue, how do I refresh the claims? Does anyone have it working, the above doesnt work.

PR #1627, opened 2018-07-16
Ignore jetbrains files

coquer:


Bogardo:
@jycr753 These should be saved in your global `.gitignore` file.

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #1628, opened 2018-07-16
Add missing throw exception

coquer:
- Changes to return types to match method signature
- Add missing throw exception


tymondesigns:
Thanks but I think it's the docblocks that need to change here

PR #1629, opened 2018-07-18
Fixed wrong auth() function in case of default guard not set to API

phucnguyenvn:
If the default guard not set to "api" we will get an errors:
+ access_token = return "true" instead of token
+ error cannot detect factory() function


shirshak55 (approved):

pulkitjalan:
the docs indicate the default guard should be updated.

PR #1635, opened 2018-07-25
Fixing bug in jwt.refresh middleware

M0H314-Hasani:
Hi , i saw a semi bug about "jwt.refresh" middleware that when i use it token refresh but decode method decode Previous Token !!
if i was wrong please forgive me because i am a beginner and please  give me solution.
thanks.


tymondesigns:
Thanks but I don't think this is necessary

PR #1643, opened 2018-08-07
Fix doc installation

gnovaro:
Rename commands. Rename files.


shcode:
You are using different branch. That documentation is supposed for version `1.x.x`

PR #1680, opened 2018-10-05
Add information for a setup with multiple guards

Christophvh:
It is possible in laravel to specify multiple guards, the docs should reflect how to specify the guard type if the api guard is not set as the default.

auth('api')->login(); will trigger the custom jwt api guard for example.


tymondesigns:
👍

PR #1682, opened 2018-10-06
Create JWTGenerateCommand.php

### LINKED ISSUE #1298: Laravel 5.5. jwt:generate issue
Hi all, 
I came upon one issue when I try to run `php artisan jwt:generate` I get the Reflection Exception saying "[ReflectionException]  Method Tymon\JWTAuth\Commands\JWTGenerateCommand::handle() does not exist"

Both Facades and service providers are entered properly and they are valid.

Also for people who have a struggle to make it work I can advise one solution. Go to **JWTGenerateCommand.php** file located in **vendor/tymon/src/Commands** and paste this part of code ` public function handle() {
        $this->fire();    
    }`

I know this is not an elegant solution, but it works. I hope this might help until official fix arrive.

Have a nice day.

eng-ahmednasser:
Fix Command  jwt:generate in Laravel 5.5

fix #1298

### DIFF
--- a/src/Commands/JWTGenerateCommand.php
+++ b/src/Commands/JWTGenerateCommand.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of jwt-auth.
+ *
+ * (c) Sean Tymon <tymon148@gmail.com>
+ *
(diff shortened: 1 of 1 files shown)


tymondesigns:
this already exists

PR #1701, opened 2018-10-18
Add getNewToken() method to BaseMiddleware

porozhnyy:
Hello! For my project, after renew the token, you need to receive it in an middleware. How to make it easier, I did not think, it looks strange. Can ideas be better?


porozhnyy:
Sorry, I did not notice that `tymon.jwt` is singletone :(

Issue #1246, opened 2017-07-11
how to test token JWT with phpunit in  Laravel


alihossein:
HI.
I want test login API in Laravel but instead of return 200 or 401, test return 302 always. 

> $info = ['email'=>'test@gmail.com','password'=>'123456'];
> $response = $this->json('POST', '/login', $info) ->assertStatus(200);

what is problem ?? 

thanks.

emersonbroga:
Same problem here.

When I make a request from a browser or postman it works, but when I make a request from phpunit, the `$headers` property inside the `JWTAuth.php` on the method `parseToken` doesn't have the authorization key. (ONLY ON TEST ENV, when running phpunit).

![JWT](http://s3.emerson.link/prints/2017-07-30-hss4qgopoxhga4lm8j2b.jpg "JWT")

emersonbroga:
Found this, `https://stackoverflow.com/questions/31309098/laravel-testcase-not-sending-authorization-headers-jwt-token`, not a definitive solution, but maybe it helps.

mehrdad-shokri:
Same problem here.  
UPDATE:  
I have found some mysterious behavior in my system, it might be specific for only my system but gonna give more info since any body else might encounter this type of problem:  

the way I generated token in my tests for a specific user was like as follows:  

     public function admin_can_get_all_degrees()
    {
        $token=JWTAuth::fromUser($this->adminUser);
        $this->get($this->DEGREES_ROUTE, ['Authorization' => "Bearer $token"])->assertStatus(200);
    }

This was throwing ``401`` status code for me.  tried getting token from the end point I had specified and it worked!  
The way I'm authenticating a user in my controller is as follows:  

    if (Auth::once(compact('email', 'password')))
        {
            $user = Auth::user();

                $token = JWTAuth::fromUser($user);
                return $this->item(new Token($token), new TokenTransformer());

        }  

Is this a bug(problem) related to tymon's jwt or I was generating token wrong?

alihossein:
thanks @mehrdaad  @emersonbroga

jgrossi:
@emersonbroga @alihossein I'm having the same issue, but I'd like to use the `$this->actingAs($user, 'api')` method, but it's not working. 

For now I created a `AttachJwtToken` trait, so inside the test class I just `use AttachJwtToken` and all calls send the JWT token for me. The trait code is here: https://gist.github.com/jgrossi/4b1364e20418eca3ca937e70550c1823.

[...]

PR #1704, opened 2018-10-27
Make testing easier with the JwtAuthentication trait

jgrossi:
> This PR is a simple solution for #1246.

### The Problem

By default, Laravel has the `actingAs()` method for tests, allowing you to log in a specific user before calling a route, for example.

When dealing with JWT the process is a bit different, because there's no session and the JWT must be sent in a request header.  😐

### The Solution

When using this package for the fist time, my first solution was overriding the `call()` method, adding a header to every call I was doing when testing. After digging some code I found an easier solution, just using a handful `withHeader()` method Laravel has.  🤔

This PR add a simple trait that overrides the `actingAs()` method when testing with Laravel, sending the `Bearer` header with the correct JWT in it. So, if you need to call some endpoint using JWT just use this trait and the `actingAs()` method:

```php
use JwtAuthentication;

public function test_the_header_is_set()
{
    $user = factory(User::class)->create();

    $this->actingAs($user)
        ->json('GET', '/foo')
        ->assertStatus(200);
}
```

And thanks for this amazing package! 👏💪🎉

### DIFF
--- a/src/Claims/JwtAuthentication.php
+++ b/src/Claims/JwtAuthentication.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of jwt-auth.
+ *
+ * (c) Sean Tymon <tymon148@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Tymon\JWTAuth\Claims;
+
+use Illuminate\Contracts\Auth\Authenticatable;
+use Tymon\JWTAuth\Facades\JWTAuth;
+
+trait JwtAuthentication
+{
+    /**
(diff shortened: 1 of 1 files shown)


cosmini:
Thanks for the solution. I'm currently writing tests for my API and I was wondering if there's a better way to login.
When can we get this feature released?

v1r0x:
This sounds great! Does this also handle several requests in one test? My current solution is to set the token using `withHeader` and have a `updateToken($response)` which parses the token returned from the request to use for the next request.

tymondesigns:
Thank you for suggesting this, but the `JWTAuth` class is deprecated (and will be removed in the next major increment) in favour using Laravel's Auth Guard. And as a result I believe the existing laravel methods work out of the box when using the jwt auth guard

PR #1700, opened 2018-10-17
Update JWTGenerateCommand.php

garam-park:
solved    ReflectionException  : Method Tymon\JWTAuth\Commands\JWTGenerateCommand::handle() does not exist

PR #1723, opened 2018-12-08
Fix Laravel Command

ChienDevIT:
There seems to be some confusion. Laravel calls handle () not fire ().

https://laravel.com/docs/5.7/artisan#command-structure

### DIFF
--- a/src/Commands/JWTGenerateCommand.php
+++ b/src/Commands/JWTGenerateCommand.php
@@ -44,11 +44,16 @@ class JWTGenerateCommand extends Command
             return $this->line('<comment>'.$key.'</comment>');
         }
 
-        $path = config_path('jwt.php');
-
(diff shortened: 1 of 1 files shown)


ChienDevIT:
A better algorithm. It is written to the .env file, not the config/jwt.php file.

https://github.com/laravel/framework/blob/5.7/src/Illuminate/Foundation/Console/KeyGenerateCommand.php#L72

tymondesigns:
See https://github.com/tymondesigns/jwt-auth/pull/1700#issuecomment-430688062

ChienDevIT:
If it is not used for laravel version 5.5 or later. I think you should set up composer.json to not support laravel 5.6 and 5.7.

Hope 1.0 will come with many improvements I hope.
Thanks,

PR #1751, opened 2019-02-12
Update LICENSE

phroggyy:
The year is not required, and having to update this every year seems pointless.


tymondesigns:
fair point `¯\_(ツ)_/¯`

PR #1756, opened 2019-02-16
Renamed all auth() calls to Auth::guard()

codekeyz:
Having followed the documentation word by word, i was always getting this error 
```php
Call to undefined function auth()
```
I managed to fix it by replacing all auth() function calls with a call to Auth::guard() using the Auth Facade 
```php
use Illuminate\Support\Facades\Auth;
```
and it worked.


tymondesigns:
`auth()` is part of the global laravel helpers

Issue #1665, opened 2018-09-13
IssuedAt Claim maxRefreshPeriod expiration detection error


sykorax:
### IssuedAt.php

Just noticed that at [this](https://github.com/tymondesigns/jwt-auth/blob/be03e7f12ba985f1474a84e0fa0b48787bb4f056/src/Claims/IssuedAt.php#L55) line of IssuedAt.php claim the predicate should be negated.

I've just set the **JWT_MAX_REFRESH_PERIOD** environment variable to some positive integer differs from null and during the unit tests it always throws expired token exception.

### Your environment
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.7.3
| Package version   | 2.0.0-dev
| PHP version       | 7.2.9

### Steps to reproduce

```php
$this->actingAs($teacher, 'api.teacher');
Auth::guard('api.teacher')->login($teacher);
```

and set **JWT_MAX_REFRESH_PERIOD** to **60**

### Expected behaviour
Not throwing an expiration exception.

### Actual behaviour
Throwing token expired exception all the time.

pulkitjalan:
Not enough steps to reproduce, feel free to reopen if its still an issue.

PR #1793, opened 2019-04-02
Fix: maxRefreshPeriod cause token expiration issue

ashutoshSce:
Fixed Issue #1665 

Check condition should be in past instead of Future.


pulkitjalan:
See comment on issue: https://github.com/tymondesigns/jwt-auth/issues/1665#issuecomment-726719160

PR #1809, opened 2019-05-14
Additional information of setup

matheussevan:
Adds configuration information as a service provider and middleware in the kernel.


tymondesigns:
I would not recommend this approach. The JWTAuth class will be deprecated in the next major release. Laravel's custom auth guard mechanism is what should be used instead as indicated already

Issue #886, opened 2016-10-29
Listening for Auth events


wemersonjanuario:
How about event listening in JWT-Auth guard? I need to listen for attemp, auttenticated and others jwt auth events like this https://laravel.com/docs/5.3/authentication#events

Using jwt-auth package I can't listen for theses events. Help?

ahmadbadpey:
I have same problem. can you find any solution ?

jubairsaidi:
The solution I've come up with (I'm not a fan of doing it this way though) is fire the events in the auth controller on the different stages of auth (attempt, login success, logout). but jwt-auth really should fire events.

wemersonjanuario:
Any official JWT Event? still need a built in solution

nivolk:
+++

chrispage1:
Same here. I was just looking at setting up authentication logging but putting it directly through the controllers isn't ideal 👎

fgilio:
Same here. Ended up doing it manually `event(new Illuminate\Auth\Events\Logout($user));`

nguyenphuocnhatthanh:
same issue...maybe we are missing function for handle it

Mapteg34:
+++

pgyf:
```php
event(new \Illuminate\Auth\Events\Login('api', Auth(‘api’)->user(), false));
```
I triggered it manually. I can't get the current guard. I can only write it dead

fabioborella-smash:
+++

leon0399:
I'm developing a library, for my own project, that relies on those events, so I'm going to make a PR adding those events

kamasuPaul:
> 
> 
> ```
> event(new \Illuminate\Auth\Events\Login('api', Auth(‘api’)->user(), false));
> ```
> 
> I triggered it manually. I can't get the current guard. I can only write it dead

For now am using this, until PR to fire events is merged by @tymondesigns

leon0399:
> > 
> > 
> > ```
> > event(new \Illuminate\Auth\Events\Login('api', Auth(‘api’)->user(), false));
> > ```
> > 
> > I triggered it manually. I can't get the current guard. I can only write it dead
> 
> For now am using this, until PR to fire events is merged by @tymondesigns 

I think it will never be merged... Project is literally dead. Better look to forks like https://github.com/PHP-Open-Source-Saver/jwt-auth and try to create PR there

leon0399:
[...]

PR #1813, opened 2019-05-19
Created events to dispatch in JWTGuard

GabrielDeveloper:
PR to #886

### DIFF
--- a/src/Events/JWTInvalidate.php
+++ b/src/Events/JWTInvalidate.php
@@ -1,7 +1,15 @@
 <?php
 
-namespace Tymon\JWTAuth\Events;
+/*
+ * This file is part of jwt-auth.
+ *
+ * (c) Sean Tymon <tymon148@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
 
+namespace Tymon\JWTAuth\Events;
 
 class JWTInvalidate
 {
@@ -22,6 +30,7 @@ class JWTInvalidate
      *
      * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
      * @param  \Tymon\JWTAuth\Token|string  $token
+     *
      * @return void
      */
     public function __construct($user, $token)
--- a/src/Events/JWTInvalidate.php
+++ b/src/Events/JWTInvalidate.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Tymon\JWTAuth\Events;
+
+
+class JWTInvalidate
+{
+    /**
+     * The authenticated user.
+     *
+     * @var \Illuminate\Contracts\Auth\Authenticatable
+     */
+    public $user;
+
+    /**
+     * @var \Tymon\JWTAuth\Token
+     */
+    public $token;
+
+    /**
+     * Create a new event instance.
+     *
+     * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
+     * @param  \Tymon\JWTAuth\Token|string  $token
+     * @return void
+     */
+    public function __construct($user, $token)
+    {
+        $this->token = $token;
+        $this->user = $user;
+    }
+}
--- a/src/Events/JWTInvalidate.php
+++ b/src/Events/JWTInvalidate.php
@@ -11,31 +11,6 @@
 
 namespace Tymon\JWTAuth\Events;
 
-class JWTInvalidate
+class JWTInvalidate extends AbstractJWTEvent
 {
-    /**
-     * The authenticated user.
-     *
-     * @var \Illuminate\Contracts\Auth\Authenticatable
-     */
-    public $user;
-
-    /**
-     * @var \Tymon\JWTAuth\Token
-     */
-    public $token;
-
-    /**
-     * Create a new event instance.
-     *
-     * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
-     * @param  \Tymon\JWTAuth\Token|string  $token
-     *
-     * @return void
-     */
-    public function __construct($user, $token)
-    {
-        $this->token = $token;
-        $this->user = $user;
-    }
 }
--- a/src/JWTGuard.php
+++ b/src/JWTGuard.php
@@ -13,10 +13,16 @@ declare(strict_types=1);
 
 namespace Tymon\JWTAuth;
 
+use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Http\Request;
 use Illuminate\Auth\GuardHelpers;
 use Tymon\JWTAuth\Claims\Subject;
 use Illuminate\Contracts\Auth\Guard;
+use Tymon\JWTAuth\Events\JWTAttempt;
+use Tymon\JWTAuth\Events\JWTInvalidate;
+use Tymon\JWTAuth\Events\JWTLogin;
+use Tymon\JWTAuth\Events\JWTLogout;
+use Tymon\JWTAuth\Events\JWTRefresh;
 use Tymon\JWTAuth\Http\TokenResponse;
 use Tymon\JWTAuth\Contracts\JWTSubject;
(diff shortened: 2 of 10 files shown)


GabrielDeveloper:
@tymondesigns Could you see this PR please? Thank you so much!

src/JWTGuard.php
```diff
@@ -121,6 +135,10 @@ public function attempt(array $credentials = [], bool $login = true)
         $this->lastAttempted = $user = $this->provider->retrieveByCredentials($credentials);
 
         if ($this->hasValidCredentials($user, $credentials)) {
+            $this->events->dispatch(
+                new JWTAttempt($user)
+            );
```
tymondesigns: I think this should be dispatched regardless of valid credentials

src/Events/JWTInvalidate.php
```diff
@@ -0,0 +1,41 @@
...
+     * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
+     * @param  \Tymon\JWTAuth\Token|string  $token
+     *
+     * @return void
+     */
+    public function __construct($user, $token)
+    {
+        $this->token = $token;
+        $this->user = $user;
+    }
```
tymondesigns: All of these events should extend a common abstract class containing the constructor and properties

tymondesigns (changes requested):
Thanks for this, see my comments

src/JWTGuard.php
GabrielDeveloper: I agree, ready!

src/Events/JWTInvalidate.php
GabrielDeveloper: Ready, thank you!

GabrielDeveloper:
@tymondesigns all requested changes are ready, thank one more time!

dhcmega:
Hi, will this then trigger the Login event? How can I install this patch?
In my composer json I have: `"tymon/jwt-auth": "^1.0"`
thanks!

GabrielDeveloper:
@dhcmega, yes.
This PR will implement the Login Event, but has not yet been merged.

When this PR is merged, you should use the 2.0.x-dev tag while 2.0 is not ready.

@tymondesigns Could you see the yours request changes, please? Thanks.

dhcmega:
Hi @GabrielDeveloper thanks for your reply.
Is 2.0-dev production ready? If that is the case, I will upgrade now to 2.
(it might sound obvious, but there are plenty of dev packages being used for prod)
Thanks!

tymondesigns:
Please do not use 2.0 in production

GabrielDeveloper:
Up =)

tymondesigns:
Thanks 👍

cubitworx:
Hi @tymondesigns Is there a roadmap for this package? Any idea when the expected release date for 2.x will be? I need access to these events. Alternatively, any chance this set of changes can be applied to 1.x? I don't mind doing a PR. Thanks in advance

Update: In hindsight, I have just gone through this implementation & it doesn't cover all of the authentication events (and I am not even sure why JWT specific events were created as opposed to using the Laravel defaults) so I'll have to go with a custom abstraction. Thanks anyway

lucianobargmann:
I cannot agree more with you @cubitworx - WHY arent we just throwing Laravel events? What is the reason for creating custom events?

Seriously, I spent a freaking week trying to figure out why my events were not being fired. And yes, let's back port this to 1.x. 

None of the other options are feasible, once I do not have a UI on my project. I will probably fork this and change the 5 lines needed.

francoism90:
@GabrielDeveloper Is it possible to merge this in `1.0.0-rc.*` as well and simple add the auth events?

GabrielDeveloper:
@francoism90, @cubitworx I created a PR to add these events on 1.0.* version: #1887 

Let's wait @tymondesigns to accept! Thanks!

GabrielDeveloper:
@lucianobargmann the events were created because we can have 2 ways to authentication on our system.

For example, we can have the default authentication to log in the system normally and we can have too an API with JWT auth to interact with mobile apps.

So I think that is a good idea we have our own events to manipulate the way that we want

lucianobargmann:
To be honest with you, I forked and solved my problem. Thanks for all the
work! Let me know if I can be of service.

Em qua, 9 de out de 2019 18:18, Gabriel Gonçalves Rodrigues <
notifications@github.com> escreveu:

nguyenphuocnhatthanh:
Tks for PR

cubitworx:
Thanks for the explanation @GabrielDeveloper but I would have at least expected the package to have Laravel events triggered in addition to your events (if need be); since you are targeting Laravel & it is an auth replacement for which Laravel has implemented a standard to work from (albeit not to your liking). The events that you have created don't really make sense to me (maybe lack of documentation) or cover all the events I would need / expect.

I ended up overloading the methods of the package I needed to in order to trigger the events.

PS. I haven't checked the PR & don't know if anything has changed from the implementation I was originally commenting on.

keyboardSmasher:
There's no need for custom events, you can just call
```
event(new Login('jwt', $user, false));
```
from `JWTGuard::login` for instance.

Then in your listener you'd have
```
$event->guard === 'jwt'
```

cubitworx:
@keyboardSmasher Having to trigger the events yourself is tantamount to creating custom events... you're trying to solve the problem of the library not fire the default auth events that anyone using Laravel would expect to listen for.

The point of using Laravel defined events is so that you end up with less vendor lock-in, as is the case when needing to use custom events for auth.

In most instances you don't care what guard triggered the event, you simply want to action something when that event happens, so this code is arguably hardly ever needed because it adds to vendor lock-in:
```
$event->guard === 'jwt'
```

keyboardSmasher:
@cubitworx
What I did was create a new class:
```
use Tymon\JWTAuth\JWTGuard;
use Illuminate\Auth\Events\Login;
use Tymon\JWTAuth\Contracts\JWTSubject;

class ExtendedJWTGuard extends JWTGuard
{
    public function login(JWTSubject $user)
    {
        $token = $this->jwt->fromUser($user);
        $this->setToken($token)->setUser($user);
        event(new Login('api', $user, false));

        return $token;
    }
}
```
Then in `AuthServiceProvider`:
```
    public function boot()
    {
        $this->app['auth']->extend('jwt', function ($app, $name, array $config) {
            $guard = new ExtendedJWTGuard(
                $app['tymon.jwt'],
                $app['auth']->createUserProvider($config['provider']),
                $app['request']
            );

            $app->refresh('request', $guard, 'setRequest');

            return $guard;
        });
```
Now I can listen for Laravel's `Login` event and differentiate between `web` and `api` in my event listener.
```
$event->guard === 'api'
```
I don't know...it works for me.

caryxiao:
> @cubitworx
> What I did was create a new class:
> 
> ```
> use Tymon\JWTAuth\JWTGuard;
> use Illuminate\Auth\Events\Login;
> use Tymon\JWTAuth\Contracts\JWTSubject;
> 
> class ExtendedJWTGuard extends JWTGuard
> {
>     public function login(JWTSubject $user)
>     {
>         $token = $this->jwt->fromUser($user);
>         $this->setToken($token)->setUser($user);
>         event(new Login('api', $user, false));
> 
>         return $token;
>     }
> }
> ```
> 
> Then in `AuthServiceProvider`:
> 
> ```
>     public function boot()
>     {
>         $this->app['auth']->extend('jwt', function ($app, $name, array $config) {
>             $guard = new ExtendedJWTGuard(
>                 $app['tymon.jwt'],
>                 $app['auth']->createUserProvider($config['provider']),
>                 $app['request']
>             );
> 
>             $app->refresh('request', $guard, 'setRequest');
> 
>             return $guard;
>         });
> ```
> 
> Now I can listen for Laravel's `Login` event and differentiate between `web` and `api` in my event listener.
> 
> ```
> $event->guard === 'api'
> ```
> 
> I don't know...it works for me.

I think it works. 😀

PR #1844, opened 2019-07-09
[0.5] Laravel 5.8 Compatibility Fixes

mbabker:
If there's interest in a new 0.5 release with "proper" support for Laravel 5.8, then this branch should have all the needed changes (and if there's no interest, no big deal, I can use this on a project until migrating to 1.0).

Changes include:

- Using `Illuminate\Contracts\Events\Dispatcher::dispatch()` when available, falling back to the deprecated `Illuminate\Contracts\Events\Dispatcher::fire()`
- Avoiding use of the deprecated array and string helper functions in favor of the methods from `Illuminate\Support\Arr` and `Illuminate\Support\Str`
- Adapting `Tymon\JWTAuth\Providers\Storage\IlluminateCacheAdapter` to account for the changes in how the cache TTL is calculated
- Allowing Carbon 2 to be used (on a cursory check it looks like none of the [B/C breaks](https://carbon.nesbot.com/docs/#api-carbon-2) apply here)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-develop": "0.5-dev"
+            "dev-master": "0.5-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "mockery/mockery": "0.9.*",
         "illuminate/auth": "~5.0",
         "illuminate/database": "~5.0",
-        "illuminate/console" : "~5.0"
+        "illuminate/console" : "~5.0",
+        "illuminate/cache" : "~5.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "illuminate/support": "~5.0",
         "illuminate/http": "~5.0",
         "namshi/jose": "^5.0 || ^7.0",
-        "nesbot/carbon": "~1.0"
+        "nesbot/carbon": "~1.0 || ~2.0"
     },
     "require-dev": {
         "phpunit/phpunit": "4.*",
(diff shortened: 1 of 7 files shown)


andrerosseto:
Good. Now, can i use 0.5 in laravel 5.8? No Carbon issues? But, i using 1.0 rc4 now...

mfn:
Shouldn't 0.5 just die?

I recently upgraded to the RC and they are fine, aren't they?

Also 0.x lacks proper support for auth guards which is essential if you want to use mutiple auths.

It feels the project would more benefit from a final stable 1.x release.

Is there  a reason to depend on 0.5 for you?

mbabker:
Haven't upgraded yet, that's what it boils down to (and we have other ongoing projects for the main client using this that are taking priority over "rewrite the JWT integration").  As this was the lone blocker for me to finish the Laravel 5.8 update, and as others seem to be having similar issues, I did my part to make it usable for those who need it.  Nothing more, nothing less.

Hazzard13:
As long as no proper upgrade guide exists to migrate to 1.0 (and there's only preview releases atm), it seems reasonable to allow users to continue using jwt-auth 0.5.* on 5.8. I know I'd certainly like to see this merged!

coquer:
Why are you targeting master? from what I can see there is a develop change to support illuminate `6`. and this is not a hotfix is a big change, why not use 2.0?

mbabker:
> Why are you targeting master?

Because master still contains the 0.5 release code, which as explained in the initial post is where this change is targeted.  And as further explained as the thread as panned out, this was blocking upgrading a live Laravel app from 5.7 to 5.8, there is not yet a stable 1.0 release, and there is not yet a migration guide to move from 0.5 to 1.0.  So instead of letting my app sit on Laravel 5.7 with this as the single blocker for upgrade, I did the work necessary for 5.8 compatibility and offered it back upstream and to other users of this package in case anyone was in a similar situation as me.  That's all there is to it.

PR #1858, opened 2019-08-15
Allow user to disable signature verification

wangfei800:
Not sure why some unrelated changes are in the PR, please review this. 

Inside a secured network, multiple services want to forward an incoming JWT token, while signature validation is ignored, here a new config variable 'verifySignature' is added.


src/Providers/JWT/Lcobucci.php
```diff
@@ -133,7 +133,7 @@ public function decode($token)
             throw new TokenInvalidException('Could not decode token: '.$e->getMessage(), $e->getCode(), $e);
         }
 
-        if (! $jwt->verify($this->signer, $this->getVerificationKey())) {
+        if ($this->laravel['config']['jwt.verifySignature'] && !$jwt->verify($this->signer, $this->getVerificationKey())) {
```
mfn: This is not the correct way to do this: configuration options are injected via the constructor, see \Tymon\JWTAuth\Providers\AbstractServiceProvider::registerLcobucciProvider

mfn (changes requested):
> Not sure why some unrelated changes are in the PR

Welp, you made the PR so you should make sure they aren't in it.

Please **only** include your _actual_ changes and fix the StyleCI violations

The changes you made are only for the `Lcobucci` adapter but you left out the `Namshi` adapter. Even though the latter is legacy, IMO the feature parity must be given.

----

Personally I'm not sure if encouraging disabling token verification is a smart move.

Can't you just catch this exception and "carry on" in your application?

src/Providers/JWT/Lcobucci.php
wangfei800: Thanks for your review, I have fixed this. 
This feature is really useful for us, as we forward the token internally among services, while we do not want to do signature verification as it had been done.

mfn (commented):
> This feature is really useful for us, as we forward the token internally among services, while we do not want to do signature verification as it had been done.

IMHO (really "h"umble ) opinion, you have a special case whose characteristics should not be suggested to be endorsed by default: being able to very easily disable the signature verification.

To me this feels almost on par with the biggest mistake of JWT itself, allowing a `none` algorithm.

My suggestion stands: catch the exception and more the code carry one.
Alternative you can explore to simply directly use the logic from `\Tymon\JWTAuth\Providers\JWT\Lcobucci::decode` in your code: parse the token, but don't verify it.

tymondesigns:
I do not want to allow this to be configurable in the config, nor do I see the benefit of doing so. 

If any part of your application needs to decode and/or trust the token payload, then the signature should always be verified, otherwise what's the point in using a jwt

PR #1864, opened 2019-09-02
Remove unsupported php versions from travis

specialtactics:
Removed PHP versions no longer supported in any way from travis testing
https://www.php.net/supported-versions.php

I was thinking potentially to update the composer.json to reflect that in requirements, but maybe there are still people for whatever reason running it unsupported. Let me know if that should be updated too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,32 +20,6 @@ matrix:
       dist: trusty
       env: LARAVEL='5.4.*' XDEBUG=1
       group: edge
-    - php: 5.5.9
-      env: LARAVEL='5.1.*'
-    - php: 5.5.9
-      env: LARAVEL='5.2.*'
-    - php: 5.5
-      env: LARAVEL='5.1.*'
-    - php: 5.5
-      env: LARAVEL='5.2.*'
-    - php: 5.6
-      env: LARAVEL='5.1.*'
-    - php: 5.6
-      env: LARAVEL='5.2.*'
-    - php: 5.6
-      env: LARAVEL='5.3.*'
-    - php: 5.6
-      env: LARAVEL='5.4.*'
-    - php: 7.0
-      env: LARAVEL='5.1.*'
-    - php: 7.0
-      env: LARAVEL='5.2.*'
-    - php: 7.0
-      env: LARAVEL='5.3.*'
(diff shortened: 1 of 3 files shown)


specialtactics:
There was a failing test in php 7.3 (due to php warning). I found out what's happening, basically there was a bit of code in one test which was using the wrong method, but PHP versions earlier than 7.3 did not say anything, and now recently it will cause a warning.

So I fixed that test.

joelwmale (approved):

.travis.yml
```diff
@@ -20,32 +20,6 @@ matrix:
       dist: trusty
       env: LARAVEL='5.4.*' XDEBUG=1
       group: edge
-    - php: 5.5.9
```
joaorobertopb, line 23: This package allows installations from **PHP version 5.5.9**, so it is important to run the tests on this version.

See: https://github.com/tymondesigns/jwt-auth/blob/6e9457185ef4e4ff52eb54a73a64be58aa46847b/composer.json#L26

joaorobertopb:
Hey @specialtactics , maybe specifying the distro in travis matrix would solve the problem.

See: https://docs.travis-ci.com/user/languages/php/#php-54x---55x-support-is-available-on-precise-and-trusty-only

specialtactics:
@joaorobertopb I am not sure what problem you refer to ? The reason I am doing this is because those PHP versions are unmaintained, so it is unreasonable for them to be supported by package developers, if even the PHP team themselves do not support them.

.travis.yml
specialtactics: I guess my suggestion is that it **shouldn't** allow those installations, on the basis that, that version of PHP hasn't been supported for many years.

joaorobertopb: Okay, but this requires updating the **PHP constraint** in the `composer.json` file.

Remove **^5.5.9** from:
https://github.com/tymondesigns/jwt-auth/blob/6e9457185ef4e4ff52eb54a73a64be58aa46847b/composer.json#L26

tymondesigns:
Thanks but I have actually resolved some most of the travis issues with this now.
Version 1 of this package will continue to allow php 5. Next version is 7.2 and above

PR #1926, opened 2020-01-22
Fixed issue with AuthHeaders parser stripping trailing hyphens from tokens

seguer:
We ran into an intermittent issue when using this library where sometimes a token (that appeared to be valid as far as we could tell) would be rendered invalid or unparseable by the library.

After some debugging, I determined it was from the AuthHeaders parser unintentionally stripping trailing hyphens from a header (eg. `Bearer foobar--` would become the token `foobar`).

On our end I added a bandaid that generates tokens without trailing hyphens, but this PR should serve to fix the issue entirely.

Note that the parse method went from using regex to simply grabbing everything in the string after the prefix (and the space separator).

### DIFF
--- a/src/Http/Parser/AuthHeaders.php
+++ b/src/Http/Parser/AuthHeaders.php
@@ -54,9 +54,9 @@ class AuthHeaders implements ParserContract
         $header = $request->headers->get($this->header) ?: $this->fromAltHeaders($request);
 
         if ($header) {
-            $start = strlen($this->prefix) + 1; // +1 to include the space separator
+            $start = strlen($this->prefix);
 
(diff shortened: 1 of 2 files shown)


specialtactics:
Not sure why those tests are failing, looks like a blip.

One thing however I will request you change, in your code you are assuming there is always exactly 1 space before the token, however current implementation allows arbitrary number of spaces.

seguer:
Good call on supporting multiple spaces, I'll rework that.

specialtactics:
Excellent work !

specialtactics (approved):

tymondesigns:
Good catch!

PR #1927, opened 2020-01-22
Remove PHP versions without support nor security fix support

specialtactics:
I just noticed there are a lot of really old PHP versions in the testing, I think they are prone to cause issues and incompatibilities, and they shouldn't really be encouraged to be used as they haven't been supported for a long time.

For reference:
https://www.php.net/supported-versions.php


.gitignore
```diff
@@ -2,3 +2,4 @@
 composer.lock
 build
 coverage.xml
+.idea
```
mfn, line 5: This shouldn't be in here. Put your IDEs meta folder in your global gitignore, please.

specialtactics: What's the issue with it being here, @mfn?

mfn: It doesn't make sense to have an ever expanding list of all the editors/IDEs in each code repository people work on. Next one wil add `.vscode`, then `.sublime` or `.eclipse` etc.

These should be in each developers global ignore.

specialtactics: What's wrong with that? There aren't really that many IDEs, I think it's fairly standard to have that in .gitignore.

tymondesigns:
Going to cleanup the php requirements soon, will also be moving from travis to github actions

tymondesigns:
closing as moved to github actions

PR #2051, opened 2020-10-29
Add more config for parsers

lanphan:
Hi,
This PR provide a convenient way for user to customize parameters for Auth Header parser (header + prefix).
In addition, I did group all configurations for parsers under main group 'parsers'.

All testcases are passed.


eschricker:
Hi, 

PR #2065, opened 2020-11-27
chore(ci): Move to GH actions and remove (most) EOL php versions

tymondesigns:
- Move to Github actions
- Remove support for EOL php versions https://www.php.net/supported-versions.php
- Migrate to phpunit v9
- Adds deprecation docblocks to classes that have been removed in v2


mfn:
Noice! ❤️

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v1
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, soap, intl, gd, gmp, exif, iconv, imagick
```
mfn: Hmm, do we really _need_ all of them?

I think only the absolute necessary ones should be required (and in turn match `require` from composer.json)?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+          coverage: ${{ matrix.coverage }}
+          ini-values: memory_limit=-1
+          tools: composer:v2
```
mfn: This can be removed, as v2 is already the latest stable default.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: "Cache dependencies"
+        uses: "actions/cache@v2"
+        with:
+          path: |
+            ~/.composer/cache
+            vendor
```
mfn: It's not advised to cache vendor directly, this can create hard to track problems.

In fact, setup-php has the most recommended way, please see https://github.com/shivammathur/setup-php#cache-composer-dependencies

mfn (commented):
🕺

.github/workflows/run-tests.yml
tymondesigns: fair point, will clean them up

.github/workflows/run-tests.yml
tymondesigns: Thanks for the clarification

.github/workflows/run-tests.yml
tymondesigns: :+1:

.gitattributes
```diff
@@ -9,6 +9,7 @@
 /.styleci.yml export-ignore
 /.travis.yml export-ignore
 /.github export-ignore
+/mkdocs.yml export-ignore
```
mfn: Seems like a duplicate now, the same is present on line 15?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,68 @@
...
+      - published
+
+jobs:
+  php-tests:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        php:
+          - 7.3
```
mfn: I think this misses `7.2`, at least according to the similar changes in `composer.json` => `"php": "^7.2"`

pulkitjalan (changes requested):
LGTM, just those few things @mfn has noticed.

.gitattributes
tymondesigns: doh!

Rohith255:
RefreshToken middleware is deprecated after upgrading to version 2.x and also noted some places deprecated in docs. Kindly give some solution what i need to do?

PR #2109, opened 2021-03-27
fix: should throw TokenInvalidException if received an invalid encrypted cookie token.

chrisLeeTW:
Currently if you had invalid encrypted token in cookie, and will throw DecryptException directly without handling.


eschricker:
Hi @chrisLeeTW , 

PR #2110, opened 2021-04-02
Update auth-guard.md: intro text for custom claims and TTL

JorisDebonnet:
I added two introductory texts for two items I had some difficulty with myself, just getting started with this: 
1. The "custom claims" mentioned here are only for general claims, not for user-specific claims. I was wondering, but where do I put the user-specific ones then? So I added that information. 
2. The TTL example seemed very misleading. The default TTL is 1 hour, which displays as 3600 seconds in the default 'expires_in' response field. So people new to this will probably associate the number 7200 with 2 hours. But `setTTL` sets it in minutes, not seconds. 7200 minutes translates to 5 days, which I guess is possible, but not intuitive, so I changed it and clarified that we're talking about minutes here.

### DIFF
--- a/docs/auth-guard.md
+++ b/docs/auth-guard.md
@@ -58,7 +58,7 @@ try {
 
 ```
 
-If the user is not set, then a `Tymon\JWTAuth\Exceptions\UserNotDefinedException` will be thrown
+If the user is not set, then a `Tymon\JWTAuth\Exceptions\UserNotDefinedException` will be thrown.
 
 ### logout()
 
@@ -73,7 +73,7 @@ auth()->logout(true);
 
 ### refresh()
 
-Refresh a token, which invalidates the current one
+Refresh a token, which invalidates the current one.
 
 ```php
 $newToken = auth()->refresh();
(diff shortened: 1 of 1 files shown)


JorisDebonnet:
(In all honesty, I find it very counter-intuitive that TTL is used in minutes in this package. I would highly recommend changing it to seconds in some future 2.x version.)

eschricker:
Hi @JorisDebonnet , 

JorisDebonnet:
> 
> 
> Hi @JorisDebonnet ,
> due to the inactivity of the project, we have forked the repository (https://github.com/PHP-Open-Source-Saver/jwt-auth). Currently we are going through the PR and identify which are relevant for us. It would be great if you would recreate your PR in our repository.

Hello, nice initiative! However, I am not actually using this package anymore. I moved to using `laravel/passport` for my API, instead of lumen with this). So feel free to recreate the changes from this PR into your fork.

eschricker:
Great, thank you for your fast reply. I will recreate the PR :)

PR #2111, opened 2021-04-11
Without realpath this will work

yoramdelangen:
Currently working on a project (with Laravel) that requires to be bundled in a Phar. The realpath is [not working within the Phar setup](https://www.php.net/manual/en/function.realpath.php#refsect1-function.realpath-returnvalues).
 
This change will allow to work properly and won't break for Laravel projects (it is only applying for Laravel projects).

### DIFF
--- a/src/Providers/LaravelServiceProvider.php
+++ b/src/Providers/LaravelServiceProvider.php
@@ -21,7 +21,7 @@ class LaravelServiceProvider extends AbstractServiceProvider
      */
     public function boot()
     {
-        $path = realpath(__DIR__.'/../../config/config.php');
+        $path = __DIR__.'/../../config/config.php';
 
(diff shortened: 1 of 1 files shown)


Messhias (approved):

Messhias:
> Currently working on a project (with Laravel) that requires to be bundled in a Phar. The realpath is [not working within the Phar setup](https://www.php.net/manual/en/function.realpath.php#refsect1-function.realpath-returnvalues).
> 
> This change will allow to work properly and won't break for Laravel projects (it is only applying for Laravel projects).

Hi, how are you doing? we forked the library to https://github.com/PHP-Open-Source-Saver/jwt-auth

If you want to migrate your PR to our repository, we'll keep it active as much we can and we're looking for someone to help us out to contribute.


Thank you,

Issue #2002, opened 2020-06-29
Token has been blacklisted


mr-Sepi0l:
### Subject of the issue
I was able to authenticate my user with th jwt and it work fine but as soon as i use the 'jwt.refresh' middleware, my next request answer is always `Token has been blacklisted`. 
I guess the old token must be invalidated as soon as my request is done but actually i can't even perform my request.

### Your environment
| Q                 | A
| ----------------- | ---
| Bug?              |  yes
| New Feature?      | no 
| Framework         | Laravel 
| Framework version | 7.0
| Package version   | 1.0
| PHP version       | 7.4

### Steps to reproduce
I'm using postman to test it. I followed every step of implementation in the doc and everything was working fine until i tried to refresh the token.
I just log in with such a function:

```php
 $credentials = request(['email', 'password']);

        if (!$token = auth()->attempt($credentials)) {
            return response()->json(['error' => 'Vos identifiants sont invalides'], 401);
        }

        return $this->respondWithToken($token);
```
I got the generated token in the answer which works fine in the next request as long as i don't use the ` 'jwt.refresh'      => 'Tymon\JWTAuth\Middleware\RefreshToken'` on my route. 
After my login, i set up the Authorization header for the next call (which use the middleware) and my token got blacklisted each time.

here is my route file:
```php
Route::group(['middleware' => ['jwt.refresh', 'jwt.auth', 'api']], function() {
    Route::resource('users', UsersController::class);
});
```

Edit: After a bit of search, in `Manager.php` the method refresh which is call in the middleware invalidate the current token which can maybe fixed by calling my authenticate middleware before the refresh one

Dissociable:
The same here in my project.
It happens sometimes. not always or even usually.

My project makes a jwt refresh request while loading and before making any other requests.
That sometimes gives Token is blacklisted message and usually works fine! TTL is set to 1Day and it happens several times in a day.
using https://github.com/websanova/vue-auth for my Laravel/Vue project.

Messhias:
I'm having this issue as well, I'm thinking to remove the blacklist configuration, but this is can result in a lot of security issues. 

And the maintainers of library it's seems not giving to much time for the opened PR's.

Issue #1573, opened 2018-05-20
The token has been blacklisted


dishcheng:
I tried this https://github.com/tymondesigns/jwt-auth/issues/1355, but when I refresh my expired token, I got  The token has been blacklisted,who can help me ? my larval project should be 5.5.40 now.
```
   "require": {
        "php": ">=7.0.0",
        "fideloper/proxy": "~3.3",
        "laravel/framework": "5.5.*",
        "laravel/tinker": "~1.0",
        "tymon/jwt-auth": "1.0.0-rc.2"
    },
```

core01:
Hi, show your code and axios interceptor

dishcheng:
my Middleware
```php
<?php

namespace App\Http\Middleware;

use Carbon\Carbon;
use Illuminate\Support\Facades\Cache;
use Tymon\JWTAuth\Exceptions\JWTException;
use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
use Tymon\JWTAuth\Http\Middleware\BaseMiddleware;
use Tymon\JWTAuth\Exceptions\TokenExpiredException;

class RefreshToken extends BaseMiddleware {

    /**
     * Handle an incoming request.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle($request, \Closure $next) {

        $this->checkForToken($request); // Check presence of a token.

        try {
            if (!$this->auth->parseToken()->authenticate()) { // Check user not found. Check token has expired.
                throw new UnauthorizedHttpException('jwt-auth', 'User not found');
            }
            $payload = $this->auth->manager()->getPayloadFactory()->buildClaimsCollection()->toPlainArray();
            return $next($request); // Token is valid. User logged. Response without any token.
        } catch (TokenExpiredException $t) { // Token expired. User not logged.
            $payload = $this->auth->manager()->getPayloadFactory()->buildClaimsCollection()->toPlainArray();
            $key = 'block_refresh_token_for_user_' . $payload['sub'];
            $cachedBefore = (int) Cache::has($key);
            if ($cachedBefore) { // If a token alredy was refreshed and sent to the client in the last JWT_BLACKLIST_GRACE_PERIOD seconds.
                \Auth::onceUsingId($payload['sub']); // Log the user using id.
                return $next($request); // Token expired. Response without any token because in grace period.
            }
            try {
                $newtoken = $this->auth->refresh(); // Get new token.
[...]

Issue #988, opened 2017-01-31
TokenBlacklistedException Exception not handling in Handler.php in Laravel 5.3


kushal555:
I am trying to handle the ``TokenBlacklistedException``  exception but it never comes in my condition.
it always return the default laravel error log.

 Below i mentioned my App/Exception/Handler.php code.
```php
if ($exception instanceof \Tymon\JWTAuth\Exceptions\TokenInvalidException) {
            $token = \Tymon\JWTAuth\Facades\JWTAuth::getToken();
            $newToken = \Tymon\JWTAuth\Facades\JWTAuth::refresh($token);
            return response(['error' => Lang::get('message.tokeninvalid')], 401)->headers("Authorization", $newToken);
        } else if ($exception instanceof \Tymon\JWTAuth\Exceptions\TokenExpiredException) {
            // Get the New Token

            return response(['error' => Lang::get('message.tokenexpired')], 400);
            #return response(['error' => Lang::get('adminlte_lang::message.tokenexpired')], 401);
        } 
        else if ($exception instanceof \Tymon\JWTAuth\Exceptions\TokenBlacklistedException) {
            
            return response()->json(['error' => Lang::get('message.tokenexpired')], $exception->getStatusCode());
        }
```

What it return 

```
FatalErrorException in JWTManager.php line 84:
Uncaught exception 'Tymon\JWTAuth\Exceptions\TokenBlacklistedException' with message 'The token has been blacklisted' in /opt/lampp/htdocs/Tranzlanta/vendor/tymon/jwt-auth/src/JWTManager.php:84
Stack trace:
#0 /opt/lampp/htdocs/Tranzlanta/vendor/tymon/jwt-auth/src/JWTManager.php(98): Tymon\JWTAuth\JWTManager->decode(Object(Tymon\JWTAuth\Token))
#1 /opt/lampp/htdocs/Tranzlanta/vendor/tymon/jwt-auth/src/JWTAuth.php(144): Tymon\JWTAuth\JWTManager->refresh(Object(Tymon\JWTAuth\Token))
#2 /opt/lampp/htdocs/Tranzlanta/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(237): Tymon\JWTAuth\JWTAuth->refresh(Object(Tymon\JWTAuth\Token))
[...]

Issue #457, opened 2016-01-28
Exception TokenBlacklistedException not working


juniormc:
I am using Laravel 5.1 and I a'm trying to use Exception: TokenBlacklistedException and TokenExpiredException but it does not work. Rest Using Advanced Client for Chrome, 
returns the following error.
 ![screen shot 2016-01-28 at 3 28 36 pm](https://cloud.githubusercontent.com/assets/14990543/12659393/2826a828-c5d4-11e5-8b59-f48b5c97dd49.png)
does not return the message in the catch.

try {
 
            if (! $user = JWTAuth::parseToken()->authenticate()) {
                return response()->json(['user_not_found'], 404);
            }
 
        } catch (Tymon\JWTAuth\Exceptions\TokenBlacklistedException $e) {
 
            return response()->json(['token_expired'], $e->getStatusCode());
 
        }catch (Tymon\JWTAuth\Exceptions\TokenExpiredException $e) {
 
            return response()->json(['token_expired'], $e->getStatusCode());
 
        } catch (Tymon\JWTAuth\Exceptions\TokenInvalidException $e) {
 
            return response()->json(['token_invalid'], $e->getStatusCode());
 
        } catch (Tymon\JWTAuth\Exceptions\JWTException $e) {
 
            return response()->json(['token_absent'], $e->getStatusCode());
 
        }

tdhsmith:
Are you in a namespace? You'll either need leading slashes on the exceptions to escape them to the global namespace, or else `use` them so they're available in the file.

juniormc:
Yes, I am using:

use JWTAuth;
use Tymon\JWTAuth\Exceptions\JWTException;
use Tymon\JWTAuth\Exceptions\TokenExpiredException;
use Tymon\JWTAuth\Exceptions\TokenInvalidException;
use Tymon\JWTAuth\Exceptions\TokenBlacklistedException;

I'm not sure if I will need something to set

juniormc:
Thanks for you comment.

I am use: 

use Tymon\JWTAuth\Exceptions\JWTException;
use Tymon\JWTAuth\Exceptions\TokenExpiredException;
use Tymon\JWTAuth\Exceptions\TokenInvalidException;
use Tymon\JWTAuth\Exceptions\TokenBlacklistedException;

and: 

try {
 
            if (! $user = JWTAuth::parseToken()->authenticate()) {
                return response()->json(['user_not_found'], 404);
            }
 
[...]

PR #2139, opened 2021-07-02
Created new configurations to optional TokenBlacklistedException

Messhias:
As been related on #2002, #457, #988, #1573, #983 (this is still open by the way), and in Laravel's application the TokenBlacklistedException shows just because we log out.

Even setting up the grace period to 30 still having the same issue, so I believe the TokenBlacklistedException should be optional since if someone is logged out, forced to log out, or even blacklisted for us (or by application) the exception most of the cases doesn't seem quite relevant to us to see.

This pull request it's a proposal to make this exception just be throwable by option.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -298,7 +298,17 @@ return [
         */
 
         'storage' => Tymon\JWTAuth\Providers\Storage\Illuminate::class,
+        
+        /*
+        |--------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


gekson (approved):

ramires352 (approved):

Messhias:
@tymondesigns can you give some attention to this one?

PR #2192, opened 2022-07-04
Void return

MathiasReker:
Add void return type to functions with missing or empty return statements, but priority is given to @return annotations.


specialtactics:
@MathiasReker tplease look at https://github.com/PHP-Open-Source-Saver/jwt-auth as this project is no longer maintained.

PR #2193, opened 2022-07-04
Explicit indirect variable

MathiasReker:
Add curly braces to indirect variables to make them clear to understand.

### DIFF
--- a/src/Claims/Factory.php
+++ b/src/Claims/Factory.php
@@ -103,7 +103,7 @@ class Factory
      */
     public function make($name)
     {
-        return $this->get($name, $this->$name());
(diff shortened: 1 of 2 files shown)


szepeviktor:
What tool do you use to discover these problems?

MathiasReker:
> What tool do you use to discover these problems?

https://github.com/FriendsOfPHP/PHP-CS-Fixer

szepeviktor:
jwt-auth uses StyleCI which is PHP CS Fixer-based.

By adding your rules to https://github.com/tymondesigns/jwt-auth/blob/develop/.styleci.yml it could be automated.

PR #2214, opened 2023-03-08
Upgrade lcobucci/jwt to 5.x

jameswagoner:


func0der:
So its either or for lcobucci/jwt? 4 or 5?

Issue #2205, opened 2023-01-07
A refresh token is used only once


moxesh-gandhi:
### How to refresh token `again` on a refresh token

Suppose my token TTL is 1 day & refresh token TTL is 2 days.

Now the case is what i generate token on 1st day it is working absolutely fine on 1st day. Now 2nd day the token is expired and we generate a refresh token and that refresh token also works perfectly fine. Now on the 3rd day refresh token is expired and we are trying to generate refresh token again on the 2nd day refresh token, **then we are unable to refresh it again**.

So how we refresh token on the old refresh token? 

The current solution is we are promoting a login form to user for asking his/her credentials to generate a new token. Now we have to stop promoting users because it's a bad way & bad user experience. And We don't need to change token TTL and refresh token TTL time.

### Your environment
| Q                 | A
| ----------------- | ---
| Bug?              | No
| New Feature?      | Yes
| Framework         | Laravel
| Framework version | 5.4 (or any version 8.75)
| Package version   | 0.5.12 (or any version 1.0.2)
| PHP version       | 7.4.3 (or any version 8.0.0)

### Steps to reproduce
Refresh the refresh token again.

### Expected behaviour
Need to refresh token on old refresh token. If refresh token is timed out then it will behave normally.

### Actual behaviour
Refresh token did not refreshed again & it's expired.

osamabelal:
You can try increasing the time span of the refresh token, and removing the refresh token from the stack when the user logs out.

mira-thakkar:
how are you able to refresh the token first time from the expired token, as i am getting 401 error while on refresh api first time?

Ore00:
@mira-thakkar Is this still an issue?

Ore00:
@tymondesigns @kingkero  can you close this?

Issue #2157, opened 2021-12-11
Token has expired and can no longer be refreshed


injectionsuccesfully:
I'm getting this error, I can't update the token. I think the problem is in the intermediate function (middleware). Should I protect the token update strategy?
Message
Token has expired and can no longer be refreshed
Level
ERROR
UserId
6
Exception
```{
   "class": "Tymon\\JWTAuth\\Exceptions\\TokenExpiredException",
   "message": "Token has expired and can no longer be refreshed",
   "code": 0,
   "file": "/var/www/app/vendor/tymon/jwt-auth/src/Claims/IssuedAt.php:59",
   "trace": [
       "/var/www/app/vendor/tymon/jwt-auth/src/Claims/Collection.php:61",

injectionsuccesfully:
For example, I created a token, and I can't update it the next day. Although the configuration files are configured correctly (to update and receive tokens), you can update the token within two weeks

HugoBravo:
Hello! I cannot refresh the token once it has expired. I can only do it before expiration. The response from the server is 401, not authorized.

'ttl' => env('JWT_TTL', 60),
'refresh_ttl' => env('JWT_REFRESH_TTL', 20160),

The response is 401, Unauthorized.

Issue #2116, opened 2021-04-28
Refresh token does not work if access_token is expired?


kamleshwebtech:
I am using Lumen 8.0 with this jwt-auth package. I am facing an issue that refresh_token does not regenerate access_token when access_token is expired?

Kindly share any suggestion, if you have. Thanks a lot. Look forward to hear you. Thanks.

BenceSzalai:
There is a setting called `JWT_REFRESH_TTL` or simply `refresh_ttl`, which defines how long the old token would be accepted for a refresh. Maybe check if it is configured correctly! Once this time-frame has expired, the old token is not accepted for generating a new one.

CodeNinja1337:
Related issue: https://github.com/tymondesigns/jwt-auth/issues/2056

zvermafia:
@BenceSzalai have you tried it by yourself? I mean `refresh_ttl` because it doesn't seem to work.

BenceSzalai:
Sorry, not, I've only used it with Laravel, where it worked fine. I just thought it may worth to check those settings.

zvermafia:
@BenceSzalai I use Laravel, too, but it doesn't work

BenceSzalai:
Well, I've never used Lumen myself, so I don't know what are the differences compared to regular Laravel. Maybe if you have shared some of your code and what you are trying to achieve someone may be able to help, but it seems more like a StackOverflow topic to me than an issue here, since it does not seem to be an issue with `jwt-auth`. I can generate a refreshed token by `$newToken = auth()->refresh();` and I have to assume it works for thousands of other users of the package...

zvermafia:
I just follow the doc, and set ttl = 1, refresh_ttl = 2. After the minute (when token time is expired) I'm trying to refresh the token, but it gives me 401 (sure, because the token time is expired and you can't authenticate with it, or maybe you could authenticate but you don't have authorization for all the actions except refreshing the token?!). I said ok and removed authentication from the refreshing route, now it works but doesn't refresh the refresh token expiration time, also there is not a method to get actual expiration times for the user rather than this strange one: `auth()->blacklist()->getRefreshTTL()`, why it needs to be called through `blacklist()`?!.

I think it must work when you write the code as shown in the documention

zvermafia:
And I'm not fighting with you, I just don't understand and write my thoughts...

BenceSzalai:
[...]

Issue #2056, opened 2020-11-17
Refresh token is not working.


abhishekdeshkar:
I'm not able to refresh the valid token. 

When I try to refresh the token it says,

```401 Unauthenticated```

How do I fix the issue ?

Thanks!

**Comment by bbprojectnet** [comment]
Did you use `JWT_MAX_REFRESH_PERIOD` (!== null)?

**Comment by abhishekdeshkar** [comment]
@bbprojectnet I don't see anything related to ```JWT_MAX_REFRESH_PERIOD``` in the official docs.

@tymondesigns  Please look into it.

**Comment by Mexidense** [comment]
Hello guys!

I don't know if you fix or it fits you but I've created a custom middleware in order to refresh and auth with JWT:

Middleware:
```batch
public function handle(Request $request, Closure $next)
    {
        if (is_null($request->bearerToken())) {
            return response()->json(['error' => 'Token required'], Response::HTTP_BAD_REQUEST);
        }

        try {
            JWTAuth::parseToken()->authenticate();
        } catch (JWTException $exception) {
            if ($request->url() === route('auth.refresh') && $exception instanceof TokenExpiredException) {
                try {
                    $token = JWTAuth::refresh();
                } catch (TokenBlacklistedException $exception) {
                    return response()->json(['error' => $exception->getMessage()], Response::HTTP_UNAUTHORIZED);
                }

                return AuthController::respondWithToken($token);
            }

            return response()->json(['error' => $exception->getMessage()], Response::HTTP_UNAUTHORIZED);
        }

        return $next($request);

```
Routes:
```batch
Route::group(
    ['prefix' => 'auth'],
    function () {
        Route::post('login', [AuthController::class, 'login'])
            ->name('auth.login');
        Route::post('logout', [AuthController::class, 'logout'])
            ->middleware('jwt')
            ->name('auth.logout');
        Route::post('refresh', [AuthController::class, 'refresh'])
            ->middleware('jwt')
            ->name('auth.refresh');
        Route::post('me', [AuthController::class, 'me'])
            ->middleware('jwt')
            ->name('auth.me');
    }
);
```
Refresh method from controller:
```batch
     /** @param string|bool
     * @return JsonResponse
     */
    public static function respondWithToken($token): JsonResponse
    {
        return response()->json(
            [
                'access_token' => $token,
                'token_type'   => 'bearer',
                'expires_in'   => auth()->factory()->getTTL() * 60
            ]
        );
    }

```

[...]

PR #2219, opened 2023-04-15
Fix for issue #2209 | #2157 | #2205 | #2056 | #2116

### LINKED ISSUE #2209: Cannot refresh an expired token even if the token is within the allowed refresh time
### Subject of the issue
I can refresh a non-expired token normally, but when I try to refresh an expired token, I get a 401 response with the message "Unauthenticated", even if the token is within the allowed refresh time. 

My JWT .env variables (I didn't put the JWT_SECRET in the printscreen but it's on the .env file):
imagem_2023-02-02_111036743

I saw other people with similar issues in the issues area but i cannot find any answer. Anyone have an idea how to resolve it?

### Your environment
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 9.47.0
| Package version   | dev-develop
| PHP version       | 8.1.2

### Steps to reproduce
Generate a token and wait till it expires. When expired, try to refresh token.

### Expected behaviour
Return the new token

### Actual behaviour
Return an 401 Unauthorized response with the message: "Unauthenticated".

imagem_2023-02-02_110859717

MohammadMehrabani:
I can refresh a non-expired token normally, but when I try to refresh an expired token, I get a 401 response with the message "Unauthenticated", even if the token is within the allowed refresh time. 
fixed #2209 | #2157 | #2205 | #2056 | #2116 with: $this->middleware('auth:api', ['except' => ['login', 'refresh']]);

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -123,7 +123,7 @@ class AuthController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('auth:api', ['except' => ['login']]);
+        $this->middleware('auth:api', ['except' => ['login', 'refresh']]);
     }
 
     /**


maoxuner:
https://github.com/tymondesigns/jwt-auth/blob/44982a70d0552cf408f7e6110d6e419cfac59f13/docs/quick-start.md?plain=1#L174

`Tymon\JWTAuth\Exceptions\JWTException` should be handled or it will result in a server internal exception when token  is invalid (not provided / blacklisted / out of refresh time / ...)

a example is:

```php
    public function refresh()
    {
        try {
            return $this->respondWithToken(auth()->refresh());
        } catch (\Tymon\JWTAuth\Exceptions\JWTException $exception) {
            throw new AuthenticationException($exception->getMessage());
        }
    }
```

PR #2222, opened 2023-06-12
phpunit.xml.dist updated, phpunit updated to ^10.0

mdmasudsikdar71:


.gitignore
```diff
@@ -1,4 +1,5 @@
 /vendor
+/.idea
```
mfn: 👎🏼 

the belongs into ppl global ignore

mfn (changes requested):
Did you test this locally?

There is more work to be done, there a couple warnings and deprecation from all over the place, making this not cleanly run through.

PR #2223, opened 2023-07-18
Dynamic authenticate key

aemaddin:
Sometimes you need to add a custom claim key that should be used instead of `sub` to retrieve the user by id on authenticate method and user method.

You just need to add a new .env key called `JWT_IDENTIFIER` that could be any claim key for example `uid`


battleplayer02 (approved):
yes this is much needed

PR #2230, opened 2023-10-05
Test against php 8.3

sergiy-petrov:


tymondesigns:
Added in https://github.com/tymondesigns/jwt-auth/pull/2243
Thanks for the PR though

PR #2233, opened 2023-10-25
:sparkles: Add new getter for JWT's factory

### LINKED ISSUE #2232: Method factory on JWTGuard not found
### Subject of the issue
Running static analysis such as PHPStan results in an error for using `$guard->factory()` even if `$guard instanceof JWTGuard` is true.

    'expires_in' => auth()->factory()->getTTL() * 60

this doesn't pass PHPStan with the error message

    Call to an undefined method                                                 
         Illuminate\Contracts\Auth\StatefulGuard&Tymon\JWTAuth\JWTGuard::factory().

### Your environment
| Q                 | A
| ----------------- | ---
| Bug?              | no / yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 10.29
| Package version   | 2.0.0
| PHP version       | 8.2

### Steps to reproduce
In a Controller that uses the `api:auth` middleware, using

```php
use App\Http\Controllers\Controller;
use Illuminate\Http\JsonResponse;
use Symfony\Component\Process\Exception\RuntimeException;
use Tymon\JWTAuth\JWTGuard;

class RefreshController extends Controller
{
    public function __invoke(): JsonResponse
    {
        $guard = Auth('api');
        if (! $guard instanceof JWTGuard) {
            throw new RuntimeException('Wrong guard returned.');
        }

        return response()->json([
            'token_type' => 'Bearer',
            'access_token' => $guard->refresh(),
            'expires_in' => $guard->factory()->getTTL() * 60,
        ]);
    }
}
```

### Expected behaviour
PHPStan should realize that `->factory()` exists on the `JWTGuard` (via the jwt token)

### Actual behaviour
Triggers the 

> Call to an undefined method
> Illuminate\Contracts\Auth\StatefulGuard&Tymon\JWTAuth\JWTGuard::factory().

error

kingkero:
fixes https://github.com/tymondesigns/jwt-auth/issues/2232

### DIFF
--- a/src/JWTGuard.php
+++ b/src/JWTGuard.php
@@ -352,6 +352,16 @@ class JWTGuard implements Guard
     }
 
     /**
+     * Get the token's auth factory.
+     *
+     * @return \Tymon\JWTAuth\Factory
+     */
+    public function factory()
+    {
+        return $this->jwt->factory();
+    }
+
+    /**
      * Get the last user we attempted to authenticate.
      *
      * @return \Illuminate\Contracts\Auth\Authenticatable


tymondesigns:
Thanks 👍🏻

PR #2239, opened 2024-01-21
feat: add laravel 11 support

MamadAliGit:


tymondesigns:
Thanks but there was a bit more effort required - see https://github.com/tymondesigns/jwt-auth/pull/2243

PR #2247, opened 2024-03-14
Add support for Carbon 3.X

RBFraphael:
Adds `^3.0` version tag to the `nesbot/carbon` dependency on `composer.json` file, to fit on Laravel 11's dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "illuminate/http": "^9.0|^10.0|^11.0",
         "illuminate/support": "^9.0|^10.0|^11.0",
         "lcobucci/jwt": "^4.0",
-        "nesbot/carbon": "^2.0"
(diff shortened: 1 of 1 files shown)


amir-khoshbakht:
https://github.com/briannesbitt/Carbon/blob/6b6383ae361003ec25488d0a1cde9551bb25e955/src/Carbon/Traits/Difference.php#L378

In the new version of the Carbon library, the signature of this function has changed from true to false.
this breaks blacklisting functionality.
https://github.com/tymondesigns/jwt-auth/blob/51620ebd5b68bb3ce9e66ba86bda303ae5f10f7f/src/Blacklist.php#L100
more here :
https://github.com/tymondesigns/jwt-auth/issues/2250#issue-2203964937

PR #2280, opened 2025-02-20
Add support for PHP 8.4 and Carbon 3

vbergerondev:
This PR adds support for PHP 8.4 as well as fixing a bug introduced by https://github.com/tymondesigns/jwt-auth/pull/2247.

See https://github.com/tymondesigns/jwt-auth/pull/2247#issuecomment-2017414717


tymondesigns:
Thanks!

PR #2292, opened 2025-08-26
chore: Inherit Doc

ChinaMoli:
When using VSCode with the [laravel/vs-code-extension](https://github.com/laravel/vs-code-extension) plugin installed, the plugin automatically determines and modifies the return value of Auth::user() to the corresponding model (`\App\Models\User`)

However, `\Tymon\JWTAuth\JWTGuard::user()` duplicates the return value, causing the return value modified by the plug-in to not take effect, so this PR modifies the comment of the user() method so that it inherits from the document.

#### laravel/vs-code-extension added file
```php
// vendor/_laravel_ide/_auth.php
<?php

namespace Illuminate\Contracts\Auth;

interface Guard
{
    /**
     * @return \App\Models\User|null
     */
    public function user();
}
```

#### Before modification
image

#### After modification
image

### DIFF
--- a/src/JWTGuard.php
+++ b/src/JWTGuard.php
@@ -64,9 +64,7 @@ class JWTGuard implements Guard
     }
 
     /**
-     * Get the currently authenticated user.
-     *
-     * @return \Illuminate\Contracts\Auth\Authenticatable|null
+     * {@inheritDoc}
      */
     public function user()
     {


specialtactics:
Makes sense

tymondesigns:
👍🏻

